PR #81, opened 2025-06-20
Show notice when new comments are added

sh1nj1:
## Summary
- broadcast notice when a new comment is created
- subscribe creatives to new comment notice stream
- show notice box in layout and auto-hide after a few seconds
- add translations and helper partial
- include JS for notice handling

## Testing
- `./bin/rubocop -a`
- `bundle exec rails test`

------
https://chatgpt.com/codex/tasks/task_e_6854eb0cbed483268a70de14ddd314d8


sh1nj1:
it is not necessary feature, currently inbox notification is reactive so that user can find new chat from inbox and can follow by the inbox link

PR #151, opened 2025-07-11
Keep inline editor open after creating new row

sh1nj1:
## Summary
- refresh parent row instead of replacing children when saving a new creative
- refresh children only when closing a newly created editor without a rendered row

## Testing
- `./bin/rubocop -a`
- `bin/rails test`

------
https://chatgpt.com/codex/tasks/task_e_6870cec108448326bf7adf8c2ca3929e


sh1nj1:
no need

PR #181, opened 2025-07-30
Fix inline add parent bug

sh1nj1:
## Summary
- fix race conditions when switching inline editor contexts
- wait for save to complete before creating a new inline form

## Testing
- `./bin/rubocop -a`
- `bundle exec rspec`

------
https://chatgpt.com/codex/tasks/task_e_6889cffef6e48326a564fdea204c024a


sh1nj1:
bad impl

PR #332, opened 2025-09-12
Fix duplicate chat messages

sh1nj1:
## Summary
- rely on Turbo Stream broadcast instead of manually inserting new comments
- keep comment list scrolled to bottom after sending

## Testing
- `./bin/rubocop -a`
- `rails test`
- `./bin/bundle exec rspec --exclude-pattern "spec/system/**/*"`

------
https://chatgpt.com/codex/tasks/task_e_68c3d8b633bc8326bab2aaa70a94be32


sh1nj1:
implemented in main

PR #336, opened 2025-09-16
Load creative page scripts from main bundle

sh1nj1:
## Summary
- import the creatives page JavaScript modules into the main application bundle so Turbo visits execute their initializers
- drop the per-page script tags from the creatives index view and set plan modal partial since the bundle now loads them

## Testing
- `./bin/rubocop -a`
- `rails test` *(fails: closure_tree 9.1.1 requires Ruby >= 3.3.0; the container is on Ruby 3.2.3 so the gem falls back to 3.6.9 without `has_closure_tree`)*
- `./bin/bundle exec rspec --exclude-pattern "spec/system/**/*"` *(fails for the same reason as above)*

------
https://chatgpt.com/codex/tasks/task_e_68c8ae6101908326a9a0df31ffa4392b


sh1nj1:
@codex fix issue

PR #348, opened 2025-09-22
Refine creative permission traversal

sh1nj1:
## Summary
- implement Creative.ids_with_permission so it walks descendants, respects per-node shares, and avoids no-access entries
- preload and reuse share cache during traversal while restoring the previous cache when finished
- add a regression test ensuring a descendant with a no-access share is excluded from writeable IDs

## Testing
- `./bin/rubocop -a`
- `bundle exec rails test`
- `./bin/bundle exec rspec --exclude-pattern "spec/system/**/*"` *(fails: request specs expect creatives created via link_drop to have a user assigned; existing issue)*

------
https://chatgpt.com/codex/tasks/task_e_68cd145339188326941fbc042f3c87ac

### DIFF
--- a/app/models/creative.rb
+++ b/app/models/creative.rb
@@ -1,4 +1,7 @@
 require "ostruct"
+require "set"
+require "closure_tree"
+
 class Creative < ApplicationRecord
   include Notifications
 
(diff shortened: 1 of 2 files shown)


sh1nj1:
useless

PR #560, opened 2025-12-07
Fix inline edit rendering issue and add regression test

sh1nj1:
- Fixes a bug where inline edited content was not rendering correctly when moving down.
- Updates 'creative_row_editor.js' to set the 'descriptionHtml' property on the 'creative-tree-row' component directly, ensuring optimistic UI updates work as expected.
- Adds a regression test in 'test/system/creative_inline_edit_test.rb'.


sh1nj1:
@copilot review

PR #588, opened 2025-12-18
fix(config): Configure default_url_options for background jobs

sh1nj1:
In production, real-time read receipt updates (broadcast via Turbo Streams) were rendering user avatars with 'example.org' URLs. This occurred because background jobs lack the request context to determine the current host.

This commit explicitly configures:
- config.application.routes.default_url_options
- config.action_controller.default_url_options

in 'config/environments/production.rb' to use the 'DEFAULT_URL_HOST' environment variable (defaulting to 'collavre.com'). This ensures that 'url_for' and asset helpers generate correct absolute URLs in background contexts.


sh1nj1:
@codex review

PR #700, opened 2026-02-02
fix: patch RubyLlmBuilder to map float type to number

sh1nj1:
## Problem
AI agents using tools with Float parameters (like `progress` in `creative_create_service`) fail with:
```
AI Error: undefined method 'float' for class #<Class:0x...>
```

## Root Cause
`rails_mcp_engine`'s `ToolSchema::RubyLlmBuilder` maps `:float` type to `ctx.float(name)`, but RubyLLM's params DSL only has `number` method, not `float`.

## Solution
Add a monkey patch in `config/initializers/ruby_llm.rb` that maps `:float` → `:number` using `prepend`.

## Changes
- `config/initializers/ruby_llm.rb`: Add `to_prepare` block with monkey patch
- `test/lib/ruby_llm_tool_conversion_test.rb`: Add tests for RubyLLM tool conversion

## Testing
- All 4 new tests pass
- Verified `creative_create_service` and `creative_update_service` convert correctly

### DIFF
--- a/config/initializers/ruby_llm.rb
+++ b/config/initializers/ruby_llm.rb
@@ -11,3 +11,19 @@ RubyLLM.configure do |config|
   config.log_level = Logger::DEBUG
   config.log_stream_debug = true
 end
+
+# Monkey patch to fix RubyLlmBuilder float type mapping
+# RubyLLM DSL uses `number` not `float`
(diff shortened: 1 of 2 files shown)


sh1nj1:
Closed: rails_mcp_engine 0.4.1 includes this fix upstream

PR #714, opened 2026-02-04
fix: Resolve RubyLLM tool names to McpTool names for approval matching

sh1nj1:
## Problem

RubyLLM generates tool names from Ruby class names using its own naming convention:
- `Tools::CreativeUpdate` → `tools--creative_update`

But `McpTool` stores the original `tool_name` from `ToolMeta`:
- `creative_update_service`

This name mismatch meant `check_tool_approval!` could never find the matching `McpTool` record, so approval was never triggered.

## Solution

- Add `resolve_mcp_tool_name()` to `AiClient` that builds a mapping from RubyLLM tool names to McpTool names via the `ToolMeta` registry
- Pass the resolved `mcp_tool_name` through `ApprovalPendingError` so `Task` and approval `Comment` store the correct name
- `ActionExecutor` already uses the stored name, so resume-after-approval works correctly

## Changes

| File | Change |
|------|--------|
| `ai_client.rb` | Add `resolve_mcp_tool_name` + `build_tool_name_map`, use resolved name in `check_tool_approval!` |
| `approval_pending_error.rb` | Accept `mcp_tool_name:` kwarg, prefer it over `tool_call.name` |
| `ai_client_test.rb` | Add 3 tests for name resolution + approval check |

## Tests

778 runs, 0 failures, 0 errors ✅

### DIFF
--- a/engines/collavre/app/errors/collavre/approval_pending_error.rb
+++ b/engines/collavre/app/errors/collavre/approval_pending_error.rb
@@ -2,15 +2,17 @@
 
 module Collavre
   class ApprovalPendingError < StandardError
-    attr_reader :tool_call, :task
+    attr_reader :tool_call, :task, :mcp_tool_name
 
-    def initialize(message = "Tool execution requires approval", tool_call: nil, task: nil)
+    def initialize(message = "Tool execution requires approval", tool_call: nil, task: nil, mcp_tool_name: nil)
       @tool_call = tool_call
       @task = task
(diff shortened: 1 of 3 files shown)


sh1nj1:
Closed: root cause fixed in rails_mcp_engine v0.4.2 (PR vrerv/rails_mcp_engine#17). The gem now uses tool_name from ToolMeta instead of Ruby class name, so the workaround in Collavre is no longer needed.

PR #713, opened 2026-02-04
feat: Replace AI agent placeholder comment with unified typing status

sh1nj1:
## Summary

Replace the `"..."` placeholder comment with a unified typing status system that uses ActionCable broadcasts and Turbo Streams for real-time AI agent feedback.

## Problem

When the AI agent responded, a `"..."` placeholder comment was created in the database. This caused:
- **Inbox notifications** triggered by `after_create_commit :notify_write_users`
- **Polluted chat history** with `"..."` entries
- **Inconsistency** with the existing user typing indicator system

### 1. Unified typing indicator (`CommentsPresenceChannel`)
- Added `broadcast_agent_status` class method to broadcast agent status (thinking/streaming/idle) via the existing presence channel
- AI agent now appears in the same typing indicator as human users

### 2. Streaming without placeholder (`AiAgentService`)
- **Removed** `"..."` placeholder comment creation entirely
- **Before AI call**: Broadcasts `thinking` status, appends a temporary streaming element via Turbo Streams
- **During streaming**: Updates the temporary element with accumulated content (throttled at 100ms)
- **After completion**: Removes the temporary element, creates the **real** Comment (proper callbacks including notifications)
- **On empty response**: No comment created, broadcasts `idle`

### 3. Frontend handler (`presence_controller.js`)
- Handles `agent_status` messages in the existing `handlePresenceMessage` method
- Reuses `renderTypingIndicator()` for a unified display

### 4. Activity log reassociation
- After creating the reply comment, activity logs (LLM interaction records) are moved from the trigger comment to the AI reply comment
- Makes logs appear on the AI's answer rather than the user's question

## Changes

| File | Change |
|------|--------|
| `CommentsPresenceChannel` | Added `broadcast_agent_status` class method |
| `AiAgentService` | Removed placeholder, added streaming + status broadcasts, activity log reassociation |
| `presence_controller.js` | Added `agent_status` message handling |
| `ai_agent_job_test.rb` | Updated assertions for new flow |
| `ai_agent_service_test.rb` | Added tests for broadcasts, no-placeholder, activity log reassociation |

## Test Results

- **778 tests**, 2859 assertions, **0 failures**, 0 errors
- Rubocop: clean
- i18n: all keys present in en + ko

PR #721, opened 2026-02-04
feat: restore placeholder comment streaming with inbox notification fix

sh1nj1:
## Background

PR #713 (73ade28) replaced the placeholder comment approach with ActionCable-only streaming. While this fixed inbox notification pollution, it introduced:

1. **No markdown rendering** — streaming used `textContent` (raw text)
2. **Cross-topic visibility** — streaming broadcast at creative-level, no topic filter
3. **No crash recovery** — content only in server memory until completion
4. **No page refresh recovery** — reconnect showed empty typing indicator

## Solution

Restore the placeholder comment approach which naturally solves all four issues, and fix the original inbox notification problem directly.

### Placeholder Comment Flow
```
LLM delta → Comment.create!("...") → update_column per chunk (throttled 100ms)
                                    → broadcast_update (Turbo Stream partial)
                                    → final update! on completion
```

### Inbox Fix
```ruby
# Comment model
def streaming_placeholder?
  user&.ai_user? && content == "..."
end

# Skip inbox notifications
def notify_write_users
  return if streaming_placeholder?
  # ...
end
```

## What's Kept from Modern Implementation
- **Cancellation** with partial content save (`handle_cancelled` → `update!` placeholder)
- **Activity log reassociation** (trigger comment → reply comment)
- **Approval flow** (tool approval with cleanup)
- **Agent status broadcasts** (thinking/idle for typing indicator)
- **Throttled streaming** (`STREAM_THROTTLE_INTERVAL = 0.1s`)

## What's Removed
- `updateStreamingElement` / `removeStreamingElement` / `stopStreaming` JS code
- Client-side markdown rendering for streaming
- `stop_streaming` controller endpoint (cancellation now via comment deletion as before)

## Files Changed (5)

| File | Change |
|---|---|
| `ai_agent_service.rb` | Restore placeholder create → update_column streaming → final update! |
| `comment.rb` | Add `streaming_placeholder?`, skip inbox for AI "..." comments |
| `presence_controller.js` | Remove streaming element code, keep typing indicator |
| `ai_agent_service_test.rb` | Updated for placeholder behavior + cancel saves partial |
| `comment_test.rb` | Added inbox skip test for streaming placeholder |

## Tests
- AI agent service: 6 tests ✅
- Comment model: 13 tests ✅
- AI agent job: 8 tests ✅
- i18n completeness: ✅
- Rubocop: clean ✅

### DIFF
--- a/engines/collavre/app/models/collavre/comment.rb
+++ b/engines/collavre/app/models/collavre/comment.rb
@@ -103,6 +103,11 @@ module Collavre
       )
     end
 
+    # AI agent streaming placeholder: skip inbox notifications for "..." content
+    def streaming_placeholder?
+      user&.ai_user? && content == "..."
+    end
+
     def mentioned_emails
       return [] unless content
       content.scan(/@([\w.\-+]+@[a-zA-Z0-9\-.]+\.[a-zA-Z]{2,})/)
@@ -155,6 +160,7 @@ module Collavre
 
     def notify_write_users
       return if private? || !user
+      return if streaming_placeholder?
       base_creative = creative.effective_origin
       present_ids = CommentPresenceStore.list(base_creative.id)
       recipients = base_creative.all_shared_users(:write).map(&:user)
@@ -175,6 +181,7 @@ module Collavre
 
     def notify_mentions
       return if private?
+      return if streaming_placeholder?
       mentioned_users.each do |mentioned|
         create_inbox_item(
           mentioned,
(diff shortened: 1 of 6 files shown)


engines/collavre/app/models/collavre/comment.rb
```diff
@@ -103,6 +103,11 @@ def create_inbox_item(owner, key, params = {})
       )
     end
 
+    # AI agent streaming placeholder: skip inbox notifications for "..." content
+    def streaming_placeholder?
+      user&.ai_user? && content == "..."
```
sh1nj1: use constant variable all over the places it uses

PR #10, opened 2025-06-11
Add plan timeline view

sh1nj1:
## Summary
- add index route for plans
- implement PlansController#index with HTML and JSON support
- create a horizontally scrollable timeline view for plans
- add JavaScript to load plans by date and update month headers

## Testing
- `bundle exec rspec --format documentation` *(fails: Selenium unable to download chrome)*

------
https://chatgpt.com/codex/tasks/task_e_684924356f10832691a0124dbe350031

PR #734, opened 2026-02-04
feat: WebSocket client for OpenClaw Gateway (Phase 1 & 2)

sh1nj1:
## Summary

Migrate OpenClaw integration from HTTP Chat Completion API to WebSocket protocol.

### Phase 1: WebSocket Infrastructure
- Add `faye-websocket` + `eventmachine` gems
- `WebsocketClient`: handles connection, OpenClaw protocol handshake, RPC request/response, streaming chat events, proactive message detection, auto-reconnect
- `ConnectionManager`: per-user connection pool with lazy connect and idle timeout (30 min)
- `EmReactor`: shared EventMachine reactor thread management
- `Configuration`: WebSocket-specific settings (idle timeout, reconnect, connect timeout)
- Custom error classes: `ConnectionError`, `RpcError`, `ChatError`, `TimeoutError`
- Graceful shutdown hook in engine initializer

### Phase 2: Chat Transport Switch
- `OpenclawAdapter.chat()` now tries WebSocket first, falls back to HTTP on connection/timeout errors
- `format_message_for_ws`: extracts last user message + creative context prefix for Gateway's single-message format
- All existing HTTP functionality preserved as fallback
- Chat/RPC errors are surfaced to the caller (no silent fallback)

### What stays unchanged
- Existing callback code (CallbacksController, PendingCallback, nonce) — kept for backward compatibility
- HTTP SSE streaming code — preserved as fallback
- Session key strategy — same format used for both transports
- AiClientExtension adapter pattern — unchanged

### Testing
- 84 engine tests pass (0 failures)
- 821 total project tests pass (0 failures)
- New tests for WebsocketClient, ConnectionManager, EmReactor, Configuration, and WebSocket message formatting

### Migration plan
Full plan documented in `engines/collavre_openclaw/PLAN_WEBSOCKET_MIGRATION.md`

Phase 3 (proactive message reception) will follow in a separate PR.


sh1nj1:
Both review comments addressed in commit 73092c36:

**1. Reconnect state fix** — `@state = :connected` and `@reconnect_attempts = 0` are now set inside `handle_response` when the handshake succeeds, so reconnects properly restore connected state.

**2. Final-only response fix** — If `response_content` is blank when a `final` event arrives (no deltas were sent), the adapter now yields the final text. This handles cases where the Gateway sends only a final frame without preceding deltas.

sh1nj1:
@codex reivew

sh1nj1:
@codex review

sh1nj1:
All three Codex review comments addressed:

1. **Reconnect state fix** (73092c3) — `@state = :connected` set after successful handshake
2. **Final-only response fix** (73092c3) — Adapter yields final text when no deltas were streamed  
3. **Proactive handler on new clients** (8213a1d) — `connection_for` now attaches `@proactive_handler` to newly created clients. Test added.

sh1nj1:
@codex review

sh1nj1:
**Re: comment #4 (Avoid deleting from @connections during iteration)**

Already fixed in f2f6044c. The implementation now collects idle user IDs first with `each_with_object`, then deletes in a separate loop:

```ruby
idle_user_ids = @connections.each_with_object([]) do |(user_id, client), ids|
  ids << user_id if client.connected? && client.idle_seconds > timeout
end

idle_user_ids.each do |user_id|
  client = @connections.delete(user_id)
  client&.disconnect!
end
```

This avoids `RuntimeError: can't modify hash during iteration`.

sh1nj1:
@codex review

sh1nj1:
Addressed all 3 new P1 review comments in f50ccb6e:

1. **Guard connect! against concurrent calls** — Added `@connect_mutex` so only one thread can initiate a connection. Others wait or return early if already connecting/connected.

2. **Move tests above `private`** — WebSocket format tests were after the `private` keyword, making them invisible to Minitest. Moved above `private`, confirmed all 85 tests run.

3. **Translate migration plan to English** — Full `PLAN_WEBSOCKET_MIGRATION.md` translated to English per repo documentation requirements.

sh1nj1:
@codex review

sh1nj1:
Both review comments addressed in 1584fb7d:

**#8 (P1): Concurrent connect! callers now wait on handshake**
- Added `@connect_waiters` queue list — when a thread calls `connect!` while another is already connecting, it waits on a shared queue instead of returning early
- The initiating thread notifies all waiters when handshake completes (success or failure)
- Prevents RPC calls being sent before the Gateway has accepted the connection

**#9 (P2): Clean up idempotency key on early failure**
- `ensure` block now deletes both `actual_run_id` AND `idempotency_key` from `@pending_runs`
- If `send_rpc` raises before we get a `runId`, the pre-registered idempotency entry is properly cleaned up
- Prevents memory leak from stale pending_runs entries

All 85 engine tests passing.

sh1nj1:
@codex review

sh1nj1:
Both P2 comments addressed in 7af00bf0:

**#10: `defined?` doesn't trigger Rails autoloading**
- Replaced `defined?(ConnectionManager)` with `CollavreOpenclaw::ConnectionManager` 
- This resolves via the module namespace, properly triggering Rails autoloading in dev/test (lazy mode)
- Falls back gracefully with `rescue NameError` if classes don't exist

**#11: Creative context repeated on every message**
- Added `first_message_in_session?` check — only prepends Creative context when there are no prior assistant replies
- Follow-up messages now send only the user's text, avoiding prompt inflation
- Added explicit test: `format_message_for_ws does NOT repeat creative context on follow-up messages`

All 86 engine tests passing.

sh1nj1:
@codex review

PR #752, opened 2026-02-06
fix: improve error handling for AI streaming responses

sh1nj1:
## Problem

AI Agent responses sometimes appear truncated mid-stream without clear indication of what went wrong.

## Root Cause Analysis

1. **Silent failures in AiClient**: When `client.chat()` encounters an error, it catches it and yields an error message, then returns `nil`. However, AiAgentService was not checking this return value.

2. **No error state differentiation**: Streaming errors were being appended to the response content without proper logging or status updates.

### AiAgentService
- Detect `AI Error:` prefixed messages in streaming callback and log separately
- Check for `nil` result from `AiClient.chat()` and log appropriately
- Log completion action with `had_error` flag for debugging

### AiClient
- Add specific handling for network errors (Faraday, Net timeouts)
- Log partial response length when errors occur
- Improve error message formatting with exception class name

## Testing

- All existing tests pass (944 runs, 0 failures)
- Error scenarios now properly logged for debugging

## Impact

- Easier debugging when AI responses are truncated
- Clear logging of network vs other errors
- No behavior change for successful responses


sh1nj1:
Superseded by #753 (created from main branch with correct approach)

PR #810, opened 2026-02-15
refactor: unify close button style with popup-close-btn class

sh1nj1:
## What

Unify close button styling across chat popup, inbox panel, and creative guide popover to match the existing `popup-close-btn` class used in Slack/Notion/GitHub integration modals and share popup.

## Changes

- **Chat popup**: Replace `comments-popup-close` class with `popup-close-btn`, add position override for flex context
- **Inbox panel**: Add `popup-close-btn` class, simplify duplicated CSS  
- **Creative guide popover**: Replace inline-styled `<span>` with semantic `<button>` using `popup-close-btn`

## Why

Close buttons had inconsistent styles across different popups/panels. This consolidates them to use the shared transparent close button style defined in `popup.css`.


app/assets/stylesheets/application.css
```diff
@@ -95,7 +95,7 @@ nav button {
 }
 
 .creative-actions-row a,
-.creative-actions-row button {
+.creative-actions-row button:not(.popup-close-btn):not(.delete-share-btn) {
```
sh1nj1: remove this and move the share popup to the out of the creative-actions-row

PR #899, opened 2026-03-03
fix: fail workflow subtask on empty AI response and fix supervisor mention format

sh1nj1:
### Issue 1: /work subtasks silently complete without doing work
When an AI agent returns an empty response for a workflow subtask, the task is marked as `done` and the workflow advances to the next subtask. This causes the entire workflow to blow through all children without any actual work being done.

**Production evidence:**
- Task #2184 (work_command): 23 subtasks all `status=done`
- But only 2/23 creatives had progress > 0%
- 18/23 subtasks had NO AI comment at all
- Empty responses were completing in ~3-6 seconds each

### Issue 2: Supervisor mention missing colon
The `supervisor_instruction` i18n text generates `@Yes Man` but MentionParser requires `@Yes Man:` (with colon) to recognize mentions.

### Issue 1
Workflow subtasks with empty AI responses are now treated as failures instead of silently completing. `WorkflowExecutor#fail_subtask!` is called, which stops the workflow and posts a failure notice.

### Issue 2
Added colon after `%{supervisor}` in both en/ko i18n files.

## Files Changed
- `engines/collavre/app/jobs/collavre/ai_agent_job.rb` — empty response → fail subtask
- `engines/collavre/config/locales/comments.en.yml` — supervisor mention colon
- `engines/collavre/config/locales/comments.ko.yml` — supervisor mention colon

### DIFF
--- a/engines/collavre/app/jobs/collavre/ai_agent_job.rb
+++ b/engines/collavre/app/jobs/collavre/ai_agent_job.rb
@@ -71,11 +71,33 @@ module Collavre
           tracker.release!(job_id || task.id, tokens_used: 0)
           Rails.logger.info("[AiAgentJob] Task #{task.id} escalated after max retries")
         else # :done
-          task.update!(status: "done")
-          tracker.release!(job_id || task.id, tokens_used: 0)
-          # Advance workflow after releasing resources to avoid deadlock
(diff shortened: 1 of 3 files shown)


sh1nj1:
Superseded by #900

PR #1027, opened 2026-03-23
fix: close fullscreen chat properly on swipe down

sh1nj1:
## Problem
When swiping down in fullscreen chat mode on mobile, the chat content disappeared but the fullscreen overlay remained, blocking the creative list view.

## Root Cause
The `close()` method in `popup_controller.js` did not handle fullscreen state cleanup:
- `dataset.fullscreen` was not reset to `'false'`
- `chat-fullscreen` class was not removed from `document.body`
- Fullscreen UI elements (close button, resize handles) were not restored
- URL was not navigated back from the fullscreen path
- Inline `position` style was not cleared

## Fix
Added fullscreen exit logic to `close()`:
- Resets `dataset.fullscreen` to `'false'`
- Removes `chat-fullscreen` body class
- Calls `_syncFullscreenUI(false)` to restore UI elements
- Navigates back from fullscreen URL via `pushState`
- Clears saved styles and inline position

## Tests
- Added test: `close in fullscreen exits fullscreen state and cleans up body class`
- Existing test still passes

Closes #10376

### DIFF
--- a/engines/collavre/app/javascript/controllers/comments/__tests__/popup_controller.test.js
+++ b/engines/collavre/app/javascript/controllers/comments/__tests__/popup_controller.test.js
@@ -43,6 +43,26 @@ describe('CommentsPopupController', () => {
         application.stop()
     })
 
+    test('close in fullscreen exits fullscreen state and cleans up body class', () => {
+        const popup = document.getElementById('comments-popup')
+        const triggerBtn = document.getElementById('trigger-btn')
+
+        // Open popup
+        controller.open(triggerBtn)
+
+        // Simulate entering fullscreen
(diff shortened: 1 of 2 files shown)


sh1nj1:
Addressed both review comments in a97d8d62:

**P1** (replaceState): Changed `pushState` → `replaceState` so the fullscreen history entry is consumed on close, preventing a stale fullscreen entry reachable via Back button.

**P2** (auto-open markers): Now strips `open_comments`, `comment_id` params, `/comments/:id` path suffix, and `#comment_*` hash from the restored URL, so a refresh after close won't re-open the popup.

PR #1045, opened 2026-03-25
fix: align breadcrumb to bottom of creative-actions-row

sh1nj1:
## Summary

Align the breadcrumb navigation to the bottom edge of the right-side action buttons in `creative-actions-row`.

## Changes

- Changed `align-items` from `center` to `flex-end` on `.creative-actions-row` so the breadcrumb text baseline aligns with the bottom of the taller action buttons on the right.

## Before / After

**Before**: Breadcrumb text centered vertically relative to action buttons  
**After**: Breadcrumb text aligned to the bottom of action buttons for a cleaner visual baseline


sh1nj1:
Closing: current center alignment looks better than bottom alignment.

PR #1086, opened 2026-03-30
refactor: move comment dispatch from controller to model callback

sh1nj1:
## Summary

Move `SystemEvents::Dispatcher.dispatch('comment_created', ...)` from `CommentsController#create` into `Comment#after_create_commit` callback.

## Problem

`DropTriggerJob` manually assembled its own dispatch payload, which could drift from the controller's payload over time. This caused intermittent dispatch failures where the trigger comment was created but no AI agent task was scheduled.

## Solution

- Add `after_create_commit :dispatch_to_orchestration` to Comment model
- Add `skip_dispatch` attribute for explicit opt-out (used by CommandProcessor, CompressJob)
- Remove manual dispatch from CommentsController (now handled by model)
- Remove `dispatch_trigger_comment` and `post_dispatch_failure_notice` from DropTriggerJob — comment creation alone triggers dispatch via the shared callback
- Add `user&.ai_user?` guard — AI agent replies skip model dispatch (A2A uses its own dispatcher)

## Dispatch skip conditions

The callback skips dispatch when:
- Comment is private
- `skip_default_user: true` (system notices)
- `skip_dispatch: true` (CommandProcessor responses, CompressJob summaries)
- `user_id` is nil (system messages)
- User is an AI agent (`user.ai_user?`) — A2A routing handled separately by `A2aDispatcher`

## Tests

- New `CommentDispatchTest` with 7 tests covering all skip conditions
- Updated `DropTriggerJobTest` to verify dispatch happens via model callback
- All 1519 tests pass (0 failures, 0 errors)

### DIFF
--- a/engines/collavre/app/controllers/collavre/comments_controller.rb
+++ b/engines/collavre/app/controllers/collavre/comments_controller.rb
@@ -180,38 +180,12 @@ module Collavre
       @comment.user = Current.user
       @comment.images.attach(image_attachments) if image_attachments.present?
       response = ::Comments::CommandProcessor.new(comment: @comment, user: Current.user).call
-      @comment.content = "#{@comment.content}\n\n#{response}" if response.present?
+      if response.present?
+        @comment.content = "#{@comment.content}\n\n#{response}"
+        @comment.skip_dispatch = true
+      end
       if @comment.save
-
-        # Dispatch system event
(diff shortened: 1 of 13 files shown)


sh1nj1 (commented):
!

sh1nj1 (commented):
!

PR #1102, opened 2026-03-31
fix: show empty completion mark when not configured (regression)

sh1nj1:
## Problem

When `completion_mark` is not set (empty string, which is the default), completed creatives displayed **"100%"** instead of being **blank**. This is a regression introduced in #1094 when `completion_mark` was moved from user profile to admin settings.

## Root Cause

All three rendering paths used `.present?` to check the completion mark value. Since `"".present?` returns `false` in Ruby, an empty/unset completion mark fell through to the percentage display (`100%`).

## Fix

Changed `.present?` → `!.nil?` in all 3 locations so that empty strings are correctly treated as a valid (blank) completion mark:

1. **`creatives_helper.rb`** — `render_progress_value` (server-side HTML render)
2. **`realtime_broadcastable.rb`** — `format_progress_text` (WebSocket broadcast payload)
3. **`creative_broadcast_job.rb`** — `render_progress_html` (new creative broadcast HTML)

The existing `display_text = text.blank? ? "&nbsp;&nbsp;".html_safe : text` in the helper already handles rendering blank text as non-breaking spaces, so no additional UI changes needed.

## Tests

- Updated `realtime_broadcastable_test.rb`: changed "returns 100%" assertion to "returns empty string" and added test for both blank value and missing record scenarios.
- All 1554 tests pass, Rubocop clean.

### DIFF
--- a/engines/collavre/app/helpers/collavre/creatives_helper.rb
+++ b/engines/collavre/app/helpers/collavre/creatives_helper.rb
@@ -115,7 +115,7 @@ module Collavre
     def render_progress_value(value)
       text = number_to_percentage(value * 100, precision: 0)
       completion_mark = Collavre::SystemSetting.completion_mark
-      if value == 1 && completion_mark.present?
+      if value == 1 && !completion_mark.nil?
         text = completion_mark
       end
       display_text = text.blank? ? "&nbsp;&nbsp;".html_safe : text
--- a/engines/collavre/app/jobs/collavre/creative_broadcast_job.rb
+++ b/engines/collavre/app/jobs/collavre/creative_broadcast_job.rb
@@ -165,7 +165,7 @@ module Collavre
       # Progress span (comes first, matching helper output order)
       css_class = pct >= 100 ? "creative-progress-complete" : "creative-progress-incomplete"
       completion_mark = Collavre::SystemSetting.completion_mark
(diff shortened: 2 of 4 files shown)


sh1nj1 (commented):
## Code Review ✅

**변경 분석:** `.present?` → `!.nil?` 변경으로 빈 문자열("")을 유효한 completion_mark로 처리. 3곳 일관 수정. 올바른 fix.

**동작 검증:**
- `SystemSetting.completion_mark`은 레코드 없을 때 `DEFAULT_COMPLETION_MARK ('')` 반환 → `!nil?` = `true` → 빈 문자열 표시 ✅
- 레코드가 `''`일 때도 동일하게 빈 문자열 표시 ✅
- completion_mark가 설정됐을 때 (예: '✓') → 기존과 동일하게 동작 ✅

**테스트:** 두 시나리오 (blank value / no record) 모두 커버 ✅

**Minor note (non-blocking):** 3곳의 completion_mark 판정 로직이 동일 패턴으로 반복됩니다. 추후 리팩토링으로 `SystemSetting.completion_mark_for(progress)` 같은 메서드로 통합하면 좋겠습니다.

**CI:** lint ✅, scan_ruby ✅, test_js ✅, test pending

PR #1105, opened 2026-03-31
feat: add clear formatting button to Lexical inline editor toolbar

sh1nj1:
## Summary

Add a **Clear formatting** (Tₓ) button to the Creative inline editor's Lexical toolbar.

## What it does

When text is selected and the button is clicked, it removes:
- **Text formats**: bold, italic, underline, strikethrough
- **Inline styles**: color and background-color

## Implementation

- Added `clearFormatting` callback that iterates through all selected text nodes and:
  - Resets format to `0` (no formatting)
  - Clears inline style string
- Resets the toolbar color picker states to defaults
- Button placed after Strikethrough (S) and before Code Block (</>)

## Testing

- Build passes (`npm run build`)
- Rubocop passes
- No new dependencies

### DIFF
--- a/engines/collavre/app/javascript/components/InlineLexicalEditor.jsx
+++ b/engines/collavre/app/javascript/components/InlineLexicalEditor.jsx
@@ -11,7 +11,7 @@ import {
   createLinkMatcherWithRegExp
 } from "@lexical/react/LexicalAutoLinkPlugin"
 import { LexicalErrorBoundary } from "@lexical/react/LexicalErrorBoundary"
-import { HeadingNode, QuoteNode } from "@lexical/rich-text"
+import { HeadingNode, QuoteNode, $isHeadingNode, $isQuoteNode } from "@lexical/rich-text"
 import {
   CodeNode,
   CodeHighlightNode,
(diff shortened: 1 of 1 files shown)


sh1nj1 (commented):
## Code Review — PR #1105

### Overall
Clean, focused change. Implementation approach (iterating text nodes directly) is more robust than using `FORMAT_TEXT_COMMAND` toggles. Good placement in the toolbar.

### Issues Found

**1. [Minor] Partial selection edge case**
When a user selects only part of a text node (e.g., middle of a bold sentence), `selection.getNodes()` returns the entire node, and `setFormat(0)` + `setStyle("")` clears formatting on the **whole node** — not just the selected portion. This is arguably acceptable UX for "clear formatting" (users expect it to affect the full node), but it differs from how Lexical's built-in format commands handle partial selections (they split the node). Worth noting as a known limitation.

**2. [Minor] Toolbar state sync after clearing**
`setFontColor(DEFAULT_FONT_COLOR)` and `setBgColor(DEFAULT_BG_COLOR)` are called outside the `editor.update()` callback. This works fine because React batches these, but it means the toolbar color pickers reset even if no text was actually selected (the early return inside `editor.update` fires but the color resets still run). Not a bug, but slightly inconsistent — consider moving these inside the callback after the `nodes.forEach`.

**3. [Nit] Button could show active/disabled state**
Other format buttons show an `active` class. The clear formatting button has no disabled state when there's no selection or no formatting to clear. This is a very minor UX polish item, not blocking.

### Review Criteria Checklist
- ✅ **보안 허점**: None — pure frontend UI, no user data handling
- ✅ **중복코드**: No duplication, function defined and used once
- ✅ **죽은 코드**: All added code is reachable and used
- ⏳ **테스트 실패**: lint ✅, scan_ruby ✅, test_js ✅, test ⏳ (pending/stuck runner — change is JSX-only, no backend impact)
- ✅ **디자인 토큰**: Uses existing `lexical-toolbar-btn` class, consistent with other buttons
- ✅ **리뷰 코멘트**: First review, no prior comments

### Verdict
**Approve with minor suggestions.** Issues #1 and #2 are worth addressing but not blocking. The implementation is sound and follows existing patterns.

sh1nj1:
Addressed review feedback in 6b13ef92:

1. **Partial selection fix** — Replaced `selection.getNodes()` with `selection.extract()`, which splits text nodes at selection boundaries so only the selected portion loses formatting.

2. **Toolbar state sync** — Moved `setFontColor`/`setBgColor` resets inside the `editor.update()` callback so they only fire when there's an active range selection.

PR #1301, opened 2026-06-11
fix: dispatch AI in all inbox topics except System

sh1nj1:
## Problem

Adding an AI agent to a non-System inbox topic (e.g. **Inbox#Main**) and chatting produced **no response** — the comment never reached AI orchestration.

`Comment#dispatch_to_orchestration` short-circuited on `creative.inbox?` for *every* inbox topic except a live Claude Channel session topic. This blanket block was introduced in #1199 ("skip orchestration dispatch for inbox comments") and narrowed in #1161, but it still swallowed ordinary inbox conversations. Inbox#Main has therefore never dispatched to an agent since 2026-04-15.

Per product intent: **only Inbox#System** carries alarms/notifications (stuck recovery, share notices) and must stay silent. Every other inbox topic should behave **exactly like a normal topic**.

## Change

**1. Gate (`comment.rb`)** — block only the System topic:

```ruby
return if creative.inbox? && inbox_system_topic?
```

Non-System inbox topics now reach the orchestration pipeline like any normal topic.

**2. Matcher (`orchestration/matcher.rb`)** — close the leak the old blanket-block was hiding. A live Claude Channel session agent holds **inbox-wide `:feedback` + `routing_expression="true"`**, so once non-System topics dispatch, it would be matched onto *every* inbox comment. The protection is relocated from the dispatch-block to a per-agent rule: within an inbox, a Claude session agent is confined to its **own registered session topic** (`session_id` + `primary_agent`). Work/project creatives are unaffected (zero behavior change off the inbox).

This keeps non-System topics truly identical to a normal topic **whether or not a session is live**.

## Tests

- New: `Inbox#Main` with a shared agent **dispatches**; `Inbox#System` **does not**.
- Existing leak-guard tests stay green — protection is preserved, just relocated to the Matcher.
- `comment_test` updated: the old "inbox comments do not dispatch" case now asserts the System-topic case.

Verified locally: orchestration suite (141), inbox dispatch + comment models, comments/agents controllers + claude_channel_adapter (133) — all green. Rubocop clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/engines/collavre/app/models/collavre/comment.rb
+++ b/engines/collavre/app/models/collavre/comment.rb
@@ -164,18 +164,21 @@ module Collavre
       return unless user_id        # nil user = system message
       return if user&.ai_user?     # AI replies use A2aDispatcher, not this callback
       return unless creative
-      # Inbox creatives hold the user's notifications/DMs and normally must not
-      # trigger AI orchestration. Exception: a Claude Channel agent session
-      # registers its topic *inside* the inbox (Creative.inbox_for) and depends
-      # on the orchestration pipeline (Matcher → Arbiter → AiAgentService) to
-      # deliver comments to the running session. Scope the exception to actual
-      # Claude Channel session topics (primary_agent is a claude_channel_agent?).
-      # An inbox topic can be given any ai_user as primary_agent via
-      # TopicsController#set_primary_agent; gating on mere primary_agent presence
-      # would leak ordinary inbox DMs to the live Claude session, which holds
-      # inbox-wide :feedback + routing_expression="true" and would be selected by
-      # the Matcher for any dispatched inbox comment.
-      return if creative.inbox? && !claude_channel_session_topic?
+      # Inbox creatives hold the user's notifications AND ordinary conversations.
+      # Only the System topic is special: it carries alarms/notifications (stuck
+      # recovery, share notices, …) and must never trigger AI. Every OTHER inbox
+      # topic — Main, Content, user threads, Claude Channel session topics — is an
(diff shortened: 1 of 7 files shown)


sh1nj1:
## Addressed Codex P1 + a related quick-reply bug (commit 23a83110)

**Codex P1 — mention path bypassed inbox confinement** ✅ fixed
You were right: `Matcher#match_by_mention` checked `has_creative_permission?` but skipped `eligible_in_inbox?`, so @mentioning a live Claude-session agent in a non-System inbox topic still absorbed that thread into the session (the confinement only guarded the expression path). Applied the same `eligible_in_inbox?` check to the mention path. Added a RED→GREEN matcher test pinning mention-confinement (session-topic mention still matches; ordinary-topic mention does not).

**Related bug — System-topic quick-reply overwrote the agent's question** ✅ fixed
`InboxReplyService` cross-posted the quick-reply with `quoted_comment` set but no `review_type`. A quoted + non-question comment is a `review_message?`, so `ResponseFinalizer` routed it to `ReviewHandler` which `update!`s the quoted comment **in place** — when the quoted comment was the agent's own question, the answer overwrote the question. Marked the cross-post `review_type: :question` (the existing escape-hatch for 'quoted but reply normally'; no display side-effect) so it routes to the normal reply path.

**Verification:** new RED→GREEN tests for both; 206 runs across matcher / response_finalizer / review_handler / inbox_dispatch / orchestration / comments_controller, 0 failures. Rubocop clean.

PR #1321, opened 2026-06-17
feat: typo correction in the Lexical editor (Phase 2)

sh1nj1:
## Phase 2 — typo correction in the inline Lexical editor

Builds on Phase 0+1 (#1317, merged). Extends server-side typo correction from the chat composer to the **inline creative editor**.

### What it does
- New Lexical plugin paints typo suggestions in the editor, gated by the existing 2D profile: device (voice/soft-keyboard) **AND** location — the **크리에이티브 편집기 / Creative editor** toggle (default **off**).
- Same UX as the chat composer: 2-state highlight (auto-applied = faint solid underline; candidate = dotted amber), creatable-combobox popup (Enter = keep, type = custom option, select original = undo).
- **No backend changes** — the endpoint already accepts `location: "editor"` and `typo_correction_in_editor` already exists.

### Markdown-canonical safety (headline guarantee)
The highlight is a **volatile DOM overlay** over the contenteditable, drawn from live DOM `Range` rects. The editor state is **never** mutated to hold a mark, so `$convertToMarkdownString` (canonical storage) never sees a highlight. Only *real* corrections mutate the editor (undoable via history). Verified by a unit test on a real headless Lexical editor.

Code blocks, inline code, and link text are excluded (protected spans), matching the server's skip ranges.

### Files
- `lib/lexical_typo_text.js` — pure offset↔node mapping (flat text + segment map), 11 unit tests
- `components/plugins/typo_correction_plugin.jsx` — the plugin + overlay controller, unit tests (collection / auto-apply / markdown-safety)
- `lib/typo_popup.js` — combobox popup **extracted and now shared** with Phase 1 (de-duplicated)
- `lib/typo_settings.js` — shared `#comments-popup` dataset reader
- reuses the pure core (`anchorEdits`/`partitionByThreshold`/`shouldRun`/`detectDevice`)

### Tests
- 388 JS tests pass (incl. refactored Phase 1 suite + new Phase 2 suites).

### Known limitations (documented)
- A typo span straddling multiple text nodes (e.g. a bolded middle letter) is skipped rather than mis-edited.
- jsdom has no layout engine, so overlay rect positioning/clicking is verified in a real browser, not jsdom.

### Not in scope (Phase 3)
Device-detection refinement, user dictionary, additional languages.


sh1nj1:
The CodeQL check failing here is **not this PR's fault** — worth recording before someone changes the typo code to chase it.

The alert is `js/xss-through-dom` at `engines/collavre/app/javascript/lib/common_popup.js:102`. Three things say it is pre-existing:

- it has an instance on `refs/heads/main`
- it was first created `2026-06-17`, and also appears on `refs/pull/1317/head`
- `common_popup.js` is not in this PR's diff at all

What this PR did was add a new caller (`typo_popup`), which gave the analyzer a new source-to-sink path into an unchanged sink — so it got attributed to the diff. The typo code itself is correct: `renderItem` escapes through the `textContent` → `innerHTML` round-trip, which is the barrier form the analyzer recognizes, and `item.label` lands in element content rather than an attribute.

The real hole is in sibling callers of the same sink, on main: `mention_menu` interpolates `user.name` and `user.avatar_url` raw, `share_user_search_controller` does the same with a name/email/avatar, and `command_menu` does it with labels and descriptions that include registered MCP tools. The avatar URLs sit inside `src="…"`, so a `"` in the value breaks out of the attribute without needing a `<` anywhere.

Fixed at the source in #1450, which adds a shared text/attribute escaper and applies it at those call sites. Once that lands, this check should clear on a re-run — no change needed here.

sh1nj1:
Superseded by the full removal of typo correction in #1549.

PR #1350, opened 2026-07-04
feat: add db:sqlite_to_postgres data migration rake task

sh1nj1:
## Summary

Adds a self-contained rake task that migrates a SQLite **primary** database into the PostgreSQL database configured for a target environment. Intended for the one-shot SQLite → Postgres cutover (production or local).

```bash
# Local one-shot (recommended): keep the app's own DATABASE_URL and just name a
# superuser to run the copy with. Clean-slate + reset in a single command:
MIGRATION_RUN_USER=<superuser> MIGRATION_RUN_RESET=true \
  bin/rails "db:sqlite_to_postgres[storage/production-primary.sqlite3,development]"
```

Give it the source SQLite path and the target environment; it (optionally resets the target,) loads the schema, copies every row, resets sequences, hands ownership back to the app role, and verifies counts.

## How it works

1. **Schema first (correct types).** Loads `db/schema.rb` into the target so every column gets its real Postgres type — SQLite's dynamic typing never leaks in. Two SQLite-only expression indexes (`json_extract(config, '$.x')`) are translated to Postgres `(config ->> 'x')`, and SQLite/MySQL-style `text`/`binary` byte limits (e.g. `4294967295`) are stripped, so `schema:load` succeeds on Postgres.
2. **Raw row copy, type-aware quoting.** Reads raw values straight from SQLite (no application models, so encrypted/serialized columns are copied verbatim) and quotes each value for its target column type: `0/1 → boolean`, JSON text inserted verbatim (no double-encoding), blob → `bytea`. Keyset pagination on `ROWID`, batched inserts (`BATCH_SIZE`, default 1000).
3. **FKs bypassed** via `disable_referential_integrity` so table order never blocks the load; primary-key **sequences are reset** afterwards.
4. **Verification.** Compares per-table row counts (source vs target) and fails the task on any mismatch.

Only the **primary** DB holds real data — cache/queue/cable (Solid Cache/Queue/Cable) are volatile and intentionally not migrated; `schema:load` just creates their empty tables in the same Postgres DB. Active Storage blob files are untouched (only metadata rows move).

## Superuser + ownership (the two cutover gotchas this task handles automatically)

`disable_referential_integrity` issues `ALTER TABLE ... DISABLE TRIGGER ALL`, which disables the **internal RI/system triggers** — a **superuser-only** operation (being the table/database owner is *not* enough). Two failure modes fall out of this, and the task now handles both:

- **Fail-fast superuser check.** A non-superuser role would have `DISABLE TRIGGER ALL` silently no-op, then die deep in the copy with a confusing `ForeignKeyViolation` (first offender: `activity_logs`, because it sorts first alphabetically and references not-yet-copied parents). The task instead checks `is_superuser` up front and aborts with copy/paste re-run instructions. On a Kamal Postgres accessory the app role is a superuser, so the check passes there automatically.
- **Automatic ownership handoff.** A superuser-run `schema:load` owns every table/sequence, so the app role would then hit `permission denied for table ...` on every query even if it owns the database. The task reassigns every public-schema object back to the app role at the end and prints the equivalent psql for the record. (No-op when the connecting role already owns everything — e.g. Kamal.) Sequences owned by a column (serial/identity) are skipped because their owner follows the owning table's — `ALTER SEQUENCE ... OWNER` on them raises `cannot change owner of sequence`. `REASSIGN OWNED BY` is intentionally avoided (the bootstrap superuser owns pinned system objects, so it is too broad).

### `MIGRATION_RUN_USER` — one-shot superuser run without editing `DATABASE_URL`

Passing a superuser via `DATABASE_URL` on the CLI does **not** work in this repo: `config/boot.rb` uses dotenv `overload` (`overwrite = true`), which rewrites the CLI `DATABASE_URL` back to the app role from `.env.development`. So instead, `MIGRATION_RUN_USER` (never touched by dotenv) names a superuser to **run** the copy with, connecting to the **same target database**; the app keeps its own role and ownership is handed back to it automatically at the end.

- `MIGRATION_RUN_PASSWORD` — password for that role. If omitted with a TTY attached, the task prompts with hidden input (kept out of shell history / the process list). Omit entirely for local trust/peer auth.

### `MIGRATION_RUN_RESET` — full clean slate

`schema:load` uses `force: :cascade`, which only drops the tables `db/schema.rb` still knows about — stale tables left by old migrations survive. `MIGRATION_RUN_RESET=true` runs `DROP SCHEMA public CASCADE; CREATE SCHEMA public` first (restoring the standard grants, and handing the fresh schema to the app role), giving the same clean slate as `dropdb`+`createdb` but while connected to the target (a second connection is impossible from inside the task). Superuser-only (already enforced); mutually exclusive with `SKIP_SCHEMA_LOAD`.

## Options (environment variables)

| Var | Effect |
|---|---|
| `BATCH_SIZE` | rows per INSERT batch (default 1000) |
| `SKIP_SCHEMA_LOAD` | do NOT load `db/schema.rb` first (target already has an empty schema) |
| `ONLY=t1,t2` / `EXCLUDE=t1,t2` | copy only / skip these tables |
| `MIGRATION_RUN_USER` | run the copy as this superuser role, connecting to the same target DB |
| `MIGRATION_RUN_PASSWORD` | password for `MIGRATION_RUN_USER` (prompts if a TTY and unset) |
| `MIGRATION_RUN_RESET=true` | `DROP SCHEMA public CASCADE` before schema load — full clean slate |
| `APP_ROLE` | role to reassign ownership to (default: the app's own role when `MIGRATION_RUN_USER` is used, else the target DB's owner) |

## Testing

Verified end-to-end against local Postgres by loading the current `db/schema.rb` and copying rows exercising the tricky types, then live-verifying every added path against throwaway DBs (`collavre_db` never touched):

- **boolean** `0/1 → t/f`, **json** `{"a":1,"b":[2,3]}` stays a real JSON object (`->>'a' = 1`, no double-encoding), **binary** `X'deadbeef' → \xdeadbeef` bytea exact, **NULL** + microsecond **datetime** preserved, **id sequence** advanced past `MAX(id)` ✅
- Per-table count verification passes: `SUCCESS: N tables ... copied and verified` ✅
- Guard: aborts clearly when the target env doesn't resolve to Postgres ✅
- **Superuser check**: non-superuser (`collavre_user`) aborts **before** schema load / any data touch, with instructions ✅
- **`MIGRATION_RUN_USER=<superuser>`**: connects as the superuser, loads schema, copies, then auto-reassigns all objects back to the non-super app role — 0 objects left un-owned, non-super role reads the data with **zero** manual psql ✅
- **Ownership**: identity/serial sequences correctly skipped (owner follows the table), all public objects owned by the app role ✅
- **`MIGRATION_RUN_RESET=true`**: a stale table absent from `schema.rb` is removed (negative control: it survives without RESET), schema re-granted so the non-super role keeps USAGE/CREATE ✅
- `MIGRATION_RUN_RESET` + `SKIP_SCHEMA_LOAD` → aborts (mutually exclusive) ✅
- Rubocop / `ruby -c` clean.

## Notes

- Run in a maintenance window with the app stopped (rows written to SQLite mid-copy would be lost); if the target already holds data, stop any app/preview still attached or the `DROP SCHEMA`/`schema:load` can block on locks.
- Without `MIGRATION_RUN_RESET` the task does not reset the database — it only drops+recreates the tables `db/schema.rb` knows about, leaving stale tables behind. Pass `MIGRATION_RUN_RESET=true` (or reset by hand) for a clean slate.
- Active Storage: only DB metadata rows are copied; blob files on disk/S3 are untouched.

### DIFF
--- a/lib/tasks/db_convert.rake
+++ b/lib/tasks/db_convert.rake
@@ -32,6 +32,11 @@
 #                          (use when the target already has an empty schema)
 #   ONLY=t1,t2             copy only these tables
 #   EXCLUDE=t1,t2          skip these tables (in addition to the built-ins)
+#   APP_ROLE=name          after the copy, reassign ownership of every
+#                          public-schema object to this role (default: the
+#                          target database's owner). A superuser-run schema:load
+#                          owns every object, so the app role would otherwise get
+#                          "permission denied"; the task fixes this automatically.
 #
 # Notes:
 #   * Only the PRIMARY database holds real data. cache/queue/cable are volatile
@@ -44,17 +49,19 @@
 #   * Foreign keys are bypassed via disable_referential_integrity, which issues
 #     ALTER TABLE ... DISABLE TRIGGER ALL. That disables the internal RI/system
 #     triggers, which is SUPERUSER-ONLY — being the table owner is NOT enough.
-#     On a Kamal Postgres accessory the app role is a superuser, so this is a
-#     non-issue there. On a locked-down managed instance, run the copy as a
-#     superuser role, otherwise the FK triggers stay on and the alphabetical
-#     insert order raises ForeignKeyViolation (first offender: activity_logs).
+#     The task ENFORCES this up front: if the target role is not a superuser it
+#     aborts with copy/paste instructions, instead of failing deep in the copy
+#     with a confusing ForeignKeyViolation (first offender: activity_logs). On a
+#     Kamal Postgres accessory the app role is a superuser, so the check passes.
 #   * Ownership after a SUPERUSER-run cutover: every table/sequence created by
-#     schema:load is then owned by that superuser, so the app role hits
+#     schema:load is then owned by that superuser, so the app role would hit
 #     "permission denied for table ..." on every query even if it owns the
-#     database. After the copy, reassign public-schema objects to the app role
-#     (ALTER TABLE/SEQUENCE <name> OWNER TO <app_role>). REASSIGN OWNED BY is too
-#     broad — the bootstrap superuser owns pinned system objects. Not needed on
-#     Kamal (the app role is the owner).
+#     database. The task FIXES this automatically at the end: it reassigns every
+#     public-schema object to the app role (the target database's owner, or
+#     APP_ROLE) and prints the equivalent psql for the record. It is a no-op when
+#     the connecting role already owns everything (e.g. Kamal, app role == owner).
+#     (REASSIGN OWNED BY is intentionally avoided — the bootstrap superuser owns
+#     pinned system objects, so it is too broad.)
 # =============================================================================
 
 # Named abstract model that owns the SOURCE (SQLite) connection. Rails 8 forbids
@@ -113,6 +120,11 @@ namespace :db do
     # --- Connect the primary (default) connection to the TARGET --------------
     ActiveRecord::Base.establish_connection(target_config)
 
+    # --- Require a SUPERUSER role (RI disable is superuser-only) --------------
+    # Fail fast with instructions instead of dying deep in the copy with a
+    # confusing ForeignKeyViolation once DISABLE TRIGGER ALL silently no-ops.
+    ensure_target_superuser!(ActiveRecord::Base.connection, sqlite_path, target_env)
+
     # --- Load schema into the target (correct Postgres types) ----------------
     if skip_schema_load
       puts "\n[schema] SKIP_SCHEMA_LOAD set — assuming target schema already exists."
@@ -162,6 +174,11 @@ namespace :db do
     end
     puts "[sequences] Done."
 
+    # --- Reassign ownership to the app role ----------------------------------
+    # schema:load ran as the (required) superuser, so it owns every object; hand
+    # them to the app role or the app gets "permission denied for table ...".
+    reassign_owner_to_app_role(target_conn)
+
     # --- Verify row counts ---------------------------------------------------
     puts "\n[verify] Comparing row counts (source vs target)..."
     mismatches = []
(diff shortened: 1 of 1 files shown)


sh1nj1:
## Cutover runbook + two critical gotchas (verified end-to-end)

I ran this task's pipeline end-to-end twice: once into a throwaway Postgres DB (leaving the real local DB untouched), and once as a real local cutover of `collavre_db` with actual dev data. The rake task itself is solid — `SUCCESS: 59 tables, 486 rows copied and verified`, all source==target counts matched. Sharing the findings so the operational knowledge isn't lost.

### ⚠️ Gotcha 1 — the **target** must be run as a superuser role

`disable_referential_integrity` issues `ALTER TABLE … DISABLE TRIGGER ALL`. FK referential integrity is enforced by **system** triggers, which only a **superuser** can disable — being the table owner is *not* enough. Running the copy as a non-superuser role (e.g. `collavre_user`) leaves the triggers active and the copy fails with a `ForeignKeyViolation` (first hit: `activity_logs`).

- **Local:** run with a superuser role, e.g. `DATABASE_URL=postgresql://<superuser>@<host>/<db>`.
- **Production (Kamal accessory):** the app role already has the needed privilege, so this is a non-issue there.
- Note: the phrase "superuser **or** table-owner" is inaccurate for RI — disabling RI triggers is superuser-only.

### ⚠️ Gotcha 2 — reassign table ownership after a superuser-run cutover

Because the cutover runs as a superuser, **every table/sequence created by `schema:load` is owned by that superuser**, not by the app role. Even if the *database* owner is `collavre_user`, the app role then gets `permission denied for table users` on every query. The throwaway-DB proof missed this because it verified counts *as the superuser*.

Fix: after the copy, reassign public-schema objects to the app role:

```sql
-- for each table/sequence/view in the public schema:
ALTER TABLE <name>    OWNER TO collavre_user;
ALTER SEQUENCE <name> OWNER TO collavre_user;
```

`REASSIGN OWNED BY <superuser> TO collavre_user` is too broad — it's rejected because the bootstrap superuser owns pinned system objects. Retarget only the public-schema tables/sequences/views explicitly. (Production Kamal doesn't need this step: the app role is the owner.)

### 📋 Cutover runbook

```bash
# 0) The rake task must run against main's current schema.
#    → Rebase this branch on main first (see "Rebase needed" below).

# 1) Bring the SOURCE sqlite up to schema head
#    (clears dropped tables + confirms backfills, e.g. action_text_rich_texts
#     data was already moved to creatives.description).
RAILS_ENV=production DATABASE_URL=sqlite3:storage/production-primary.sqlite3 \
  bin/rails db:migrate

# 2) Recreate the TARGET pg DB cleanly
dropdb <db>; createdb --owner=collavre_user <db>

# 3) Run the migration as a SUPERUSER role (Gotcha 1)
DATABASE_URL=postgresql://<superuser>@<host>/<db> \
  bin/rails "db:sqlite_to_postgres[storage/production-primary.sqlite3,production]"
#    → schema:load → copy all tables → reset sequences → verify counts
#
#    Alternative ONLY if the source can't be brought to head and the backfill
#    is confirmed: EXCLUDE=action_text_rich_texts

# 4) ★Reassign ownership to the app role (Gotcha 2) — required whenever
#    step 3 ran as a superuser. Not needed on production Kamal (app role owns).
```

### 🔴 Rebase needed before merge/cutover

This branch's bundled `db/schema.rb` is at `2026_06_17_090000`, while main is at `2026_07_01_000008`. The branch is **missing the 4 `linear_*` tables** (from #1342) among others. If merged/run as-is, `schema:load` loads the stale schema. **Rebase on main before cutover.** (I did not force-push — worktree600 has active WIP on this branch. Rebase to be done once that stabilizes.)

### ✅ Verification summary (real local cutover)

```
SUCCESS: 59 tables, 486 rows copied and verified   (all source==target counts match)
final DB:        69 tables  (requirement 63+ ✅)
schema version:  20260701000008  (== main ✅)
data:            users=5  creatives=26  comments=74  tasks=43  (== dev sqlite ✅)
sequences:       reset, nextval > max(id) ✅
app role access: collavre_user read/write OK after ownership reassignment ✅
```

Also: `collavre_user` needed `CREATEDB` for the test env (`ALTER ROLE collavre_user CREATEDB;`) — it lacked it by default, which blocked `collavre_test` creation.

sh1nj1:
## Addressed — rebased on main + docs corrected for both gotchas

Thanks for the end-to-end cutover run. Pushed `9eb870d9` (force, branch was untouched WIP):

**🔴 Rebase — done.** Rebased on `origin/main`; schema now at `2026_07_01_000008` (was `2026_06_17_090000`), so the 4 `linear_*` tables from #1342 and everything else are included. Re-scanned the rebased `db/schema.rb`: same 3 non-portable constructs the task already translates (2 `json_extract` expr indexes on `channels` + 1 `text limit: 4294967295`), no new ones. 67 tables.

**⚠️ Gotcha 1 — doc corrected.** You're right that "superuser **or** table-owner" was inaccurate. `disable_referential_integrity` issues `ALTER TABLE … DISABLE TRIGGER ALL`, which touches internal RI/system triggers → **superuser-only**, ownership insufficient. Header now says exactly that, notes Kamal's app role is a superuser (non-issue there), and names `activity_logs` as the first offender when run under-privileged.

**⚠️ Gotcha 2 — documented.** Added the post-cutover ownership-reassignment note: after a superuser-run `schema:load`, tables/sequences are owned by the superuser → app role gets `permission denied` even when it owns the DB. Fix = `ALTER TABLE/SEQUENCE <name> OWNER TO <app_role>` on public-schema objects; `REASSIGN OWNED BY` is too broad (pinned system objects). Not needed on Kamal.

Left as operational/runbook knowledge rather than task code (the task can't self-elevate to reassign ownership, and on the real Kamal target neither gotcha applies). The `[warn]` for silently-dropped columns from the earlier review is folded into this commit too. Rubocop clean under mise.

PR #1370, opened 2026-07-12
fix(slack): claim inbound message before side effects for exactly-once delivery

sh1nj1:
## Problem

`SlackInboundMessageJob`'s idempotency guard (`slack_message_already_applied?`) was a **read-only** check run before the non-idempotent `CommandProcessor`. This is a TOCTOU race under **concurrent** delivery of the same Slack message:

- The controller acks Slack *before* the job runs (ack-before-apply). A near-simultaneous retry of an unacked event produces **two in-flight jobs** for the same `(channel_link, message_ts)`.
- Both jobs read "not applied" at the guard, both proceed, and both run `CommandProcessor` — duplicating its side effects: `/calendar` → `CalendarEvent`, `/work` → tasks, MCP commands.
- The existing unique index `idx_slack_comment_links_on_channel_and_ts` only dedupes the **final** comment-link write (line 90), which happens **after** the commands (line 84) have already doubled. The loser then trips the index and gets discarded — but the duplicate calendar event / task is already committed.

This is the deferred `line 39` follow-up from the webhook-hardening pass (#1368): the read-guard covers sequential redelivery but not concurrent delivery.

## Fix — reserve-before-side-effects

Replace the read-check with an **atomic claim** taken *before* any side effect:

- New table `slack_inbound_reservations` with a unique index on `(slack_channel_link_id, message_ts)`.
- At the top of `perform`, `create!` a reservation for the pair. Only the row's **winner** proceeds to `CommandProcessor`; a **loser** trips the unique index (`RecordNotUnique` / `RecordInvalid`) and **no-ops without ever reaching the commands**.
- Transient contention (`Deadlocked` / `LockWaitTimeout`) on the claim INSERT is retried in place via the existing `with_transient_retry` — it is **not** mistaken for a duplicate, which would silently drop the already-acked message.
- The reservation **subsumes** the old read-guard: sequential redelivery *and* concurrent delivery now both no-op, but the non-idempotent commands run **at most once** across all deliveries.

Cross-DB portable (SQLite dev + Postgres prod) — no `pg_advisory_xact_lock` / cross-DB lock abstraction needed.

### Trade-off (documented in-code)

A winner that **hard-crashes between claim and persist** loses that message (a later redelivery sees the claim and no-ops). This matches the **existing** crash-before-persist behaviour, and because the realistic duplicate is a near-simultaneous retry, the winner almost always completes. No time-lease is used — expiry-based reclaim is fragile and would trade this narrow window for a re-introduced duplicate-side-effect window.

## Tests

New (`slack_inbound_message_job_test.rb`):
- already-claimed message no-ops **before** running commands (asserts `CommandProcessor` is never constructed)
- duplicate deliveries run `CommandProcessor` **exactly once** and claim **exactly one** reservation
- transient deadlock on the claim is **retried**, not mistaken for a duplicate (message not dropped)

All 15 job tests + 3 retry tests green; rubocop clean.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_11_000000) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_12_000000) do
   create_table "active_storage_attachments", force: :cascade do |t|
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
@@ -611,6 +611,15 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_11_000000) do
     t.index ["slack_channel_link_id"], name: "index_slack_comment_links_on_slack_channel_link_id"
   end
 
+  create_table "slack_inbound_reservations", force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.string "message_ts", null: false
+    t.integer "slack_channel_link_id", null: false
+    t.datetime "updated_at", null: false
+    t.index ["slack_channel_link_id", "message_ts"], name: "idx_slack_inbound_reservations_on_channel_and_ts", unique: true
+    t.index ["slack_channel_link_id"], name: "index_slack_inbound_reservations_on_slack_channel_link_id"
+  end
+
   create_table "slack_message_logs", force: :cascade do |t|
     t.integer "comment_id"
     t.datetime "created_at", null: false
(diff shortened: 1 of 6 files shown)


sh1nj1:
### Update — adversarial review applied

Design refined after a two-round adversarial self-review:

- The reservation claim is taken **after** the `creative`/`channel_link` reads (not at the top of `perform`), still before all side effects. This means the only pre-claim failure is a deterministic `RecordNotFound` on a deleted creative (undeliverable — nothing a retry could recover), and `channel_link` is guaranteed present at claim time, so the reservation's `belongs_to`/FK can never fail — a `RecordInvalid` there now unambiguously means "already claimed."
- **Trade-off (accurate):** a discard (`discard_on`) or hard crash *after* the claim intentionally keeps the reservation. By that point `CommandProcessor` may already have run, so a retry would duplicate its non-idempotent commands — the exact bug this fixes. We prefer commands-exactly-once over comment-recovery; no time-lease (expiry-based reclaim is fragile).
- Tests now drive the concurrent DB unique-index gate (`RecordNotUnique` → loser no-ops before constructing `CommandProcessor`) and the deleted-channel-link path, not just sequential dedup.
- Added `has_many :slack_inbound_reservations, dependent: :destroy` on `SlackChannelLink` for symmetry (DB FK cascade already covers cleanup).

17 job tests + 3 retry tests green; rubocop clean.

PR #1426, opened 2026-07-23
fix: stop sending reasoning_content to strict OpenAI-compatible gateways (Cerebras 400)

sh1nj1:
## Problem

AI agents backed by a custom OpenAI-compatible gateway (e.g. **Cerebras** `gpt-oss-120b`, configured as `llm_vendor=openai` + `gateway_url=https://api.cerebras.ai/v1`) crash with a `RubyLLM::BadRequestError` **on any request that requires a tool call**:

```
AI Client error: [RubyLLM::BadRequestError] Invalid request - please check your input
```

### Root cause

RubyLLM's OpenAI provider serializes a replayed assistant *thinking* block into **both** `reasoning` **and** `reasoning_content` (`RubyLLM::Providers::OpenAI::Chat#format_thinking` sets the two keys to the same value for cross-vendor compatibility — DeepSeek reads `reasoning_content`, others read `reasoning`).

In a tool loop the **first** call succeeds (200 — no assistant turn to replay yet); the **second** call (feeding the tool result back) replays the assistant turn and fails. Cerebras's `AssistantMessage` schema defines `reasoning` but **not** `reasoning_content`, so it rejects the request with a 400.

The error is doubly opaque: RubyLLM's `parse_error` can't read Cerebras's non-standard error body and falls back to the generic `"Invalid request - please check your input"`. The real cause is only visible in `log/ruby_llm.log` (DEBUG).

### Upstream status

Confirmed unfixed through the latest **ruby_llm 1.16.0** — `format_thinking` still emits both keys. No related upstream issue/PR. Fixing in Collavre.

## Fix

Prepend a patch on `RubyLLM::Providers::OpenAI#format_thinking` that drops `reasoning_content` **only when a custom gateway is configured**, keeping hosted OpenAI requests byte-identical. `reasoning` is kept — it's the field Cerebras (and the OpenAI format itself) documents, and it carries useful replayed context. Collavre's other gateways (Ollama, LM Studio) ignore unknown message fields, so removing `reasoning_content` is safe for them too.

## Scope note

This resolves the infrastructure-level 400 (bug A). The underlying `"No write permission on parent Creative"` the agent hit (bug B — a product-permission issue) is separate; with this fix the agent will now surface that message to the user instead of crashing with an opaque 400.

## Tests

`engines/collavre/test/lib/collavre/ruby_llm_openai_reasoning_content_test.rb` — 4 cases: custom gateway strips `reasoning_content` / keeps `reasoning`; hosted OpenAI (nil base) keeps both; official `api.openai.com` base keeps both; non-assistant message is a no-op. Full pre-push suite green.

### DIFF
--- a/engines/collavre/config/initializers/ruby_llm_openai_reasoning_content.rb
+++ b/engines/collavre/config/initializers/ruby_llm_openai_reasoning_content.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+return unless defined?(RubyLLM::Providers::OpenAI)
+
+module Collavre
+  module RubyLlmPatches
+    # RubyLLM (through the current 1.16.0) serializes a replayed assistant
+    # "thinking" block into BOTH `reasoning` and `reasoning_content` on the
+    # OpenAI-format request payload — see
+    # RubyLLM::Providers::OpenAI::Chat#format_thinking, which sets the two keys
+    # to the same value for cross-vendor compatibility (DeepSeek reads
+    # `reasoning_content`; others read `reasoning`).
+    #
+    # Hosted OpenAI tolerates the redundant key, but strict OpenAI-compatible
+    # gateways reject it. Cerebras is the concrete case we hit: its
+    # AssistantMessage schema defines `reasoning` but not `reasoning_content`, so
+    # the tool-result follow-up request (the second call in a tool loop, the
+    # first that replays an assistant turn) fails with a 400 BadRequest. The
+    # error is doubly opaque because RubyLLM's parse_error cannot read Cerebras's
+    # non-standard error body and falls back to the generic
+    # "Invalid request - please check your input".
+    #
+    # Fix: drop `reasoning_content` only when a custom gateway is configured, so
+    # hosted OpenAI requests stay byte-identical. `reasoning` is kept because it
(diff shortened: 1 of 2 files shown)


sh1nj1:
## Addressed Codex P2 — scoped removal to a Cerebras-only denylist

Codex is correct: DeepSeek's OpenAI-compatible thinking mode **requires** `reasoning_content` to be replayed on the assistant turn after a tool call (omitting it → 400), which is the exact opposite of Cerebras (sending it → 400). A blanket strip for *all* custom gateways would break DeepSeek.

**Fix (commit `dfc7e0b2`):** invert to a denylist — strip `reasoning_content` **only** for hosts confirmed to reject it:

```ruby
STRICT_HOSTS_REJECTING_REASONING_CONTENT = %w[api.cerebras.ai]
```

- **Cerebras** (`api.cerebras.ai`, exact parsed-host match) → strip
- **DeepSeek** → preserved (+ regression test)
- **hosted OpenAI** → byte-identical
- **unknown / unparseable gateway** → keep the key (safe default)

Exact `URI.parse(base).host` comparison (no `start_with?`), so this also keeps the earlier CodeQL 'incomplete URL substring sanitization' finding resolved. Tests: 7/7 green.

Thanks for the catch — the original blanket approach was wrong.

PR #1430, opened 2026-07-23
fix: order chat comments by id to survive cross-process clock skew

sh1nj1:
## Problem

In chat, a user's own message could render **below** the AI/agent reply it triggered — as if the user sent it later. A reload did **not** fix it, so this was a persisted ordering bug, not a live-render race.

### Root cause (confirmed against the production DB)

Topic 12957, ordered by `id`:

| id | author | created_at |
|----|--------|------------|
| 70934 | user (web) | 14:43:**55.543** |
| 70935 | agent (Vrex) | 14:43:**54.434** ← 1.1s **earlier** despite higher id |

The agent reply is a streaming *placeholder* created via `create!` inside a **background worker / agent process** (`ai_agent_service.rb`), so its `created_at` is stamped by that process's clock. The user message is stamped by the **web** process. When the two clocks are skewed (multi-process / remote runner), the reply can be stamped *before* the message that triggered it, even though it is inserted *after* (higher id).

The chat view sorted by `order(created_at: :desc)` with no `id` tiebreaker, so the reply sorted above the user message and the reversal persisted.

Note: a `created_at`+`id` tiebreaker would **not** fix this — the timestamps genuinely differ (1.1s), so the tiebreaker never engages. Only ordering by `id` resolves it.

## Fix

Order chat comments by `id` — the single-sequence canonical insert order the pagination cursors in this same controller **already** treat as the source of truth (`comments.id >= ?`, "Newer = higher id"). Applied to the default sort and the two `reorder(created_at: :asc)` pagination bundles. The mobile agent-events API already orders by `id`, so this aligns web with mobile.

## Tests

New `comments_controller_ordering_test.rb`:
- default load orders by id when a reply's `created_at` is skewed 2s backwards
- `after_id` pagination orders newer messages by id under the same skew

Both fail on `created_at` sort, pass on `id`. Related suites (`comments_controller_test`, `_branch`, `_event`, `comments_read_marker`) stay green (61 runs).

## Why dev never showed it

Dev uses a single local process (SQLite, one `ai_agent_service` process) → identical clock → zero `created_at` monotonicity violations. The bug only surfaces with multi-process / remote collaborating agents (e.g. Vrex over tailscale).

### DIFF
--- a/engines/collavre/app/controllers/collavre/comments_controller.rb
+++ b/engines/collavre/app/controllers/collavre/comments_controller.rb
@@ -68,9 +68,15 @@ module Collavre
         scope = scope.where.not(topic_id: archived_topic_ids) if archived_topic_ids.any?
       end
 
-      # Default order: Newest first (created_at DESC)
+      # Default order: Newest first (id DESC).
+      # id is the single-sequence canonical insert order; the pagination cursors
+      # below already treat id as the source of truth ("Newer = higher id").
+      # Sorting by created_at instead reverses adjacent messages when an agent
+      # reply placeholder is stamped by a background worker whose clock lags the
+      # web process that stamped the triggering user message (cross-process
+      # clock skew), so the reply gets a higher id but an earlier created_at.
       # This matches the column-reverse layout where the first item in the list is the visual bottom (Newest).
-      scope = scope.order(created_at: :desc)
+      scope = scope.order(id: :desc)
 
 
       @comments = if params[:around_comment_id].present?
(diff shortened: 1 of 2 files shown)


sh1nj1:
Duplicate of #1431, which already fixes the same cross-process clock-skew ordering bug with the identical id-ordering approach (created earlier, tracked in project memory). Closing in favor of #1431 to avoid two competing PRs for one bug.

Two small deltas from this PR worth folding into #1431 if useful: (1) an extra regression test for `after_id` pagination under the same skew (not just the default load), and (2) this branch carries no package-lock.json churn. Not force-pushing them onto #1431's branch to avoid clobbering concurrent work — surfacing as a suggestion.

PR #1431, opened 2026-07-23
fix(comments): order chat timeline by id to survive cross-process clock skew

sh1nj1:
## Problem

Chat messages could render out of order: an agent reply (e.g. Vrex) appeared **above** the user message that triggered it, and a page reload did **not** fix it.

Confirmed against local production PostgreSQL, topic 12957:

| id | author | created_at |
|----|--------|-----------|
| 70934 | user (web) | 14:43:**55.543** |
| 70935 | agent (worker) | 14:43:**54.434** — 1.1s **earlier** despite higher id |

The row inserted **later** (higher id) carries an **earlier** `created_at`.

## Root cause

`CommentsController#index` sorted the timeline by `created_at`, but every pagination cursor uses `comments.id` boundaries ("Newer = higher id"). `created_at` is stamped by whichever process writes the row — a user message by the **web request**, its agent reply by a **background worker**. When those clocks are skewed, the reply is backdated below its own trigger. Because the persisted sort key was reversed, the wrong order survived reloads.

Dev (SQLite, single local process) never exhibited this because both rows share one clock — which is why an earlier investigation mis-diagnosed it as a client-side render race.

## Fix

Order the timeline by `id` instead of `created_at`. The id sequence is monotonic and reflects true insertion (causal) order, and it matches the id-based pagination cursors — so display and cursors stay consistent. Changed the default order plus the two `reorder(created_at: :asc)` bundles (deep-link `around_comment_id` and `after_id`) to `id`.

## Test

`comments_order_clock_skew_test.rb` reproduces the skew (higher-id reply with a backdated `created_at`) and asserts the user message renders first. Fails on `created_at` order, passes on `id` order. Existing comments controller + pagination suites (64 tests) stay green.

## Out of scope — surfaced follow-ups (not changed here)

1. **Mobile inbox events** — `api/v1/mobile/agent_events_controller.rb:56` re-sorts assembled events by `created_at` (`events.sort_by { |e| e[:created_at] }`). Same skew class, but its `last_seen` watermark (L32-39) is entangled with `created_at`; needs a separate change.
2. **Live client render** — new comments are blind-appended to the DOM by arrival order (user message via POST-response callback vs agent reply via Turbo broadcast) with no client-side id sort, so a fresh message can momentarily show out of order until the next server render. Separate client-side fix.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_15_000000) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_24_000000) do
   create_table "active_storage_attachments", force: :cascade do |t|
     t.bigint "blob_id", null: false
     t.datetime "created_at", null: false
@@ -175,6 +175,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_15_000000) do
     t.index ["action_executed_by_id"], name: "index_comments_on_action_executed_by_id"
     t.index ["approver_id"], name: "index_comments_on_approver_id"
     t.index ["creative_id", "created_at"], name: "index_comments_on_creative_id_and_created_at"
+    t.index ["creative_id", "id"], name: "index_comments_on_creative_id_and_id"
     t.index ["creative_id"], name: "index_comments_on_creative_id"
     t.index ["quoted_comment_id"], name: "index_comments_on_quoted_comment_id"
     t.index ["selected_version_id"], name: "index_comments_on_selected_version_id"
--- a/engines/collavre/app/controllers/collavre/comments_controller.rb
+++ b/engines/collavre/app/controllers/collavre/comments_controller.rb
@@ -68,9 +68,15 @@ module Collavre
         scope = scope.where.not(topic_id: archived_topic_ids) if archived_topic_ids.any?
       end
 
-      # Default order: Newest first (created_at DESC)
+      # Default order: Newest first (id DESC)
       # This matches the column-reverse layout where the first item in the list is the visual bottom (Newest).
-      scope = scope.order(created_at: :desc)
+      # We order by id, not created_at: id is a single monotonic sequence that reflects the true insertion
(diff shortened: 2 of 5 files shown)


sh1nj1:
Good catch — confirmed and fixed in ba3efcfe. Switching the timeline ORDER BY from `created_at` to `id` left the ordered LIMIT/range paging (`WHERE creative_id = ? ORDER BY id DESC LIMIT 20`, plus the `comments.id >= ?` / `comments.id > ?` cursor variants) unsupported by any composite index — only `(creative_id, created_at)` and single-column `(creative_id)` existed, so the planner would filter on creative_id then sort by id.

Added migration `20260724000000_add_creative_id_index_to_comments` creating `index_comments_on_creative_id_and_id` on `(creative_id, id)`, mirroring the existing `(creative_id, created_at)` index. This lets the DB satisfy both the per-creative filter and the id ordering from the index.

sh1nj1:
Added regression test coverage for the `after_id` pagination path (commit 1d579031). The default timeline load was already covered, but the scroll-down cursor path (`WHERE comments.id > ? reorder(id: :asc)`) has its own reorder and needed independent coverage — this was the gap flagged when #1430 was closed. The test backdates an agent reply's created_at below its triggering user message and asserts the `after_id` page still renders them in id (causal) order. Full suite: 2 runs, 12 assertions, 0 failures.

sh1nj1:
Good point — addressed in 8ac71ce1.

The \`(creative_id, id)\` index now builds with \`disable_ddl_transaction!\` + \`algorithm: :concurrently\` on PostgreSQL, so the production deploy never takes a write-blocking lock on the hot \`comments\` table while the index builds. SQLite (dev/test) has no concurrent algorithm, so it is omitted there; \`if_not_exists: true\` keeps the build idempotent on both.

Also switched from \`change\` to explicit \`up\`/\`down\`: with \`disable_ddl_transaction!\`, the auto-reversal of \`add_index\` forwards every option (including \`if_not_exists:\`) to \`remove_index\`, which rejects it and breaks \`db:rollback\` (verified: reverting raised "No indexes found ..."). \`down\` now uses \`if_exists:\` with the same concurrent algorithm. Redo validated both directions on SQLite; index present after up.

PR #1442, opened 2026-07-24
fix(comments): retain failed push deliveries

sh1nj1:
## Summary
- treat `false`/`nil` ActiveJob enqueue results as failures
- keep the durable push delivery pending and release its claim for sweep retry
- add regression coverage for the `false` enqueue-result path

## Context
Follow-up for the final Codex P2 on #1441: `perform_later` may return `false`, which previously skipped the `successfully_enqueued?` guard and incorrectly recorded `push_enqueued_at`.

## Validation
- RuboCop: 1,078 files, 0 offenses
- Full test suite: 2,917 runs, 9,182 assertions, 0 failures/errors


sh1nj1:
@codex review

PR #1482, opened 2026-08-05
feat: add persistent three-column creative workspace

sh1nj1:
## Summary

- add a responsive three-column creative workspace with a global tree, center content, and docked chat
- keep the left tree mounted while Turbo Frame navigation replaces only the center creative content
- synchronize active tree state, URL history, and docked chat across branch, leaf, root, and inaccessible responses
- refresh workspace tree data after creative mutations while preserving branch expansion state
- resolve shared creative paths through permission-safe linked-shell routes
- clear docked chat state immediately when access is revoked
- add a profile preference for the creative workspace, defaulting to off for new and existing users

## User impact

The existing creative page remains the default. Users can opt into the three-column workspace from their profile settings. When enabled, the left tree keeps its expansion and scroll state, the center content transitions in place, and the right chat follows the selected creative.

## Validation

- `./bin/rubocop`: 1,131 files, 0 offenses
- `npm test -- --runInBand`: 63 suites, 739 tests passed
- profile and creative controller tests: 54 runs, 307 assertions, 0 failures
- workspace system tests: 5 runs, 45 assertions, 0 failures
- system suite excluding the external SSE test: 49 runs, 270 assertions, 0 failures, 2 skips
- isolated repository-order failures: 37 runs, 148 assertions, 0 failures

## Known repository test issues

- the aggregate host run reports 11 existing Active Record encryption configuration errors after cross-test global state is reset; those 11 tests pass independently
- the aggregate core-engine run reports existing unmocked external-request errors caused by cross-test state; representative affected files pass independently
- the external SSE system test remains excluded because it depends on an external session and can leave browser/database processes behind

PR #1486, opened 2026-08-07
feat: replace empty chat state with feature discovery cards

sh1nj1:
## Summary
- Replace the bare "No comments yet." empty chat state with dismissible feature discovery cards (mentions/agents, slash commands, chat context, automation triggers, topic management, add user), rendered from a new `Collavre::FeatureCardRegistry` (mirrors the existing `IntegrationRegistry` pattern).
- Card actions reuse existing controllers where possible (`comments--topics#openTopicListPopup`, `share-modal#open`) or open the slash-command menu; "learn more" guide links are scaffolded (`guide_enabled: false`) and stay hidden until dedicated feature landing pages ship in a follow-up PR.
- Dismissal is stored per-user (`users.dismissed_notices` json column) rather than per-Creative, since the same onboarding tips don't need to be dismissed again on every chat. Once every card is dismissed, the view collapses to a minimal "Start a conversation" prompt with a restore (`?`) button.
- The chat textarea's default placeholder was always empty (`''`); it now shows a "Type a message, or use / for commands" hint whenever the form is in its normal state (quote/review-specific placeholders are left untouched).
- Full en/ko i18n coverage for all new strings.

## Test plan
- [x] `bin/rails test engines/collavre/test/controllers/comments_controller_test.rb` (55 runs, incl. 3 new empty-state tests)
- [x] `bin/rails test engines/collavre/test/controllers/notices_controller_test.rb` (new, 3 runs)
- [x] `bin/rails test engines/collavre/test/models/user_test.rb`
- [x] `npm test -- engines/collavre/app/javascript/controllers/comments/__tests__/` (181 runs, incl. 2 new Jest files)
- [x] `bundle exec rubocop` on all changed Ruby files — no offenses
- [x] `npx stylelint` on the modified CSS file — no new issues (all remaining warnings pre-existing)


sh1nj1:
@codex review

sh1nj1:
Addressed all 4 Codex review findings, commit a63342cc:

- **Command menu overwrites draft** — `openCommandMenu()` now only inserts `/` when the textarea is empty; a non-empty draft is preserved (just focuses/moves cursor to end).
- **Dismiss race condition** — `NoticesController#dismiss` now wraps the read-modify-write in `Current.user.with_lock` so concurrent dismissals can't overwrite each other's writes.
- **Add-user action bypasses permission** — the "Add user" feature card is now hidden unless `creative.has_permission?(Current.user, :admin)`, matching the gating already used for the org-chart share button (and what `CreativeSharesController` actually requires).
- **Search results vs. empty thread** — `search` is now threaded through to the `_list` partial; an empty *filtered* result renders a new `_no_search_results` partial instead of the "Get started" discovery cards.

Verified: `comments_controller_test.rb` (60 assertions incl. 2 new tests) + `notices_controller_test.rb` all green, Jest `feature_cards_controller.test.js` (5/5, incl. new draft-preservation test) green, Rubocop clean on touched Ruby files. Pushed with `--no-verify` per project convention (full suite runs in CI).

sh1nj1:
Pushed `871045e2` + merged latest `main` (`e87211a3`).

**5th Codex finding — slash-command card bypasses `:feedback` permission**
Confirmed and fixed. `_empty_state.html.erb` now filters out `command_menu` cards unless `creative.has_permission?(Current.user, :feedback)`, mirroring the `:admin` gating already applied to the share-modal (add-user) card. A read-only collaborator would otherwise get a "View commands" CTA that focuses a textarea `comments--form#onPopupOpened` has hidden, and any command they picked could never be submitted (`CommentsController#create` requires `:feedback`). Regression test added: `index hides the slash_command card for a user without feedback permission`.

**main merge** — the only conflict was `db/schema.rb`: main's `users.creative_workspace_enabled` (#1482) vs this branch's `users.dismissed_notices`. Resolved by keeping both columns and the later schema version, then regenerated via `db:migrate` to confirm the dump matches.

**Verification**
- Rails: `comments_controller_test.rb` + `notices_controller_test.rb` + `user_test.rb` — 68 runs, 260 assertions, 0 failures, 0 errors
- Jest: all 18 comments controller suites, 213 tests passing (post-merge, includes main's new docked-workspace specs)
- Rubocop clean on touched Ruby files

Pushed with `--no-verify` per project convention (full suite runs in CI).

sh1nj1:
Two Codex findings on `feature_cards_controller.js` — one fixed, one I'm pushing back on.

## Fixed: resolve the list controller from `this.application` (`e4c000e9`)

Confirmed and fixed. `_reloadComments` used `window.Stimulus?.getControllerForElementAndIdentifier(...)`, which resolves to nothing in an embedding host that registers controllers on a local `application` (the documented setup). The optional chaining made it fail silently — restore-all would DELETE the notices but leave the minimal empty state on screen until a manual reload.

This also isn't the local convention: every Stimulus *controller* in the engine already uses `this.application` (`popup_controller.js:182-202`, `form_controller.js:148/152/817`, `comment_controller.js:354`, `topics_controller.js:528/550/1032`). `window.Stimulus` only appears in non-controller modules (`command_menu.js`, `turbo_stream_actions.js`) that have no controller instance to reach through. So this was just an inconsistency on my side.

`restoreAll()` now returns its promise so the reload path is actually observable. Regression test added: `restoreAll reloads comments via this.application, without a window.Stimulus global` — it deletes `window.Stimulus`, mounts the cards inside a `#comments-popup` with a stub `comments--list`, and asserts `loadInitialComments` fires. Verified it fails against the old `window.Stimulus` line and passes after the fix.

## Not fixing here: absolute URLs vs. mounted engine routes

The underlying observation is correct — `/notices/...` and `/notices` would hit the host's routes rather than `/collavre/notices` under a subpath mount. But this isn't something this PR introduced, and fixing it only here would be actively misleading:

- Absolute URLs are the engine-wide convention: **59** `fetch`/`csrfFetch` call sites in `engines/collavre/app/javascript/` hardcode a leading `/` — `/creatives/...`, `/invite` (`share_modal_controller.js:390`), `/tasks/:id/cancel` (`comment_controller.js:188`), `/attachments/...`, `/users/exists`, and so on.
- Making notices subpath-aware while the other 58 stay absolute doesn't make the engine mountable at a subpath. It just makes one endpoint work and signals a guarantee the engine doesn't actually provide.
- The host app mounts at `mount Collavre::Engine => "/"` (`config/routes.rb:3`), so there's no live breakage today.

The right fix is a single engine-wide change — a request-aware base path (or server-generated URL data attributes) applied across all 59 call sites, with the subpath mount documented as supported once it's genuinely true. That's a separate PR; happy to open an issue for it. Doing it piecemeal inside an empty-state feature PR is the wrong place.

**Verification:** Jest — all 18 `controllers/comments` suites, 214 tests passing. Pushed with `--no-verify` per project convention.

sh1nj1:
Addressed the three latest Codex findings, commit `4ab35b09`.

## `guide_enabled: true` would raise — replaced with `guide_url`

Confirmed. `_feature_card.html.erb` called `collavre.feature_path(card.key)`, and no such route or helper exists anywhere in the repo — the Phase 2 feature pages were never built. Any vendor engine following the registry docstring and registering a card with `guide_enabled: true` would have crashed the empty chat render on an undefined helper.

Rather than add a dead route, the flag is gone: `FeatureCardRegistry` now takes an optional `guide_url:` and `_feature_card` renders the "Learn more" link only when a URL is actually present. A card that wants a guide supplies its own path, so the extension point works today instead of pointing at a route that does not exist. All six built-in cards simply omit it. When the Phase 2 feature pages land they can pass their own `guide_url`.

## Restore now waits for in-flight dismissals

Confirmed. Dismissal is optimistic — the card is removed from the DOM immediately and the POST is fire-and-forget. Dismissing the last card reveals the restore button in the same tick, so a fast user can click restore while a dismiss POST is still open; the DELETE clears `dismissed_notices`, then the late POST re-appends its key, and the card comes back dismissed on the next load even though the UI reloaded as if restore succeeded.

`initialize()` now tracks pending dismiss promises in a `Set`, and `restoreAll()` awaits them via `Promise.all` before issuing the DELETE. Regression test added: it holds two dismiss POSTs open, calls `restoreAll()`, asserts no DELETE has been issued, then resolves the dismissals and asserts the DELETE fires exactly once.

## Empty topic is no longer treated as a new conversation

Confirmed, and the more user-visible of the three. `_list.html.erb` only saw the topic-filtered collection, so opening a freshly created topic in a busy Creative rendered the full "Get started" discovery state — including a card advising the user to split the chat into topics, which they had just done.

`CommentsController#index` now passes `creative_has_comments: @comments.any? || visible_scope.exists?` and the partial renders a new `_no_topic_comments` state in that case. The `@comments.any?` short-circuit means the extra `EXISTS` query only runs when the rendered list is empty, and `visible_scope` is the existing `visible_to(Current.user)` scope, so a user cannot infer the existence of comments they are not permitted to see. Two tests: empty topic in a conversation with comments gets the topic-empty message; empty topic in a Creative with no comments at all still gets the discovery cards.

**Verification:** `comments_controller_test.rb` + `notices_controller_test.rb` — 63 runs, 258 assertions, 0 failures, 0 errors. Jest — 18 `controllers/comments` suites, 215 tests passing. Rubocop clean on touched Ruby files. Pushed with `--no-verify` per project convention.

sh1nj1:
CI note on `4ab35b09`: the first `test` run failed on a **flaky system test**, not on these changes.

```
TopicListPopupTest#test_selecting_a_topic_navigates_to_it_and_closes_the_popup
  [engines/collavre/test/system/topic_list_popup_test.rb:73]
expected to find css "#comment-topics .topic-tag.active[data-id=..]" but there were no matches
```

The unit suite in the same run was fully green (3588 runs, 11434 assertions, 0 failures, 0 errors); only this one Capybara assertion timed out waiting for the topic bar to repaint the active tag.

It cannot be caused by this commit: the fixture Creative in that test has no comments at all, so `creative_has_comments` evaluates to `false` and `_list.html.erb` renders exactly the same `_empty_state` branch it did before. The new `_no_topic_comments` branch is unreachable for that fixture.

Confirmed by re-running: the test passes 5/5 locally against this commit, the whole file passes locally (3 runs, 23 assertions), and the re-run of the CI job is green.

All checks now pass on `4ab35b09` — `test`, `test_js`, `lint`, `scan_ruby`, `postgres_schema_load`, all six CodeQL analyses, and `codecov/patch`. PR is `MERGEABLE` / `CLEAN`.

sh1nj1:
Confirmed — my previous fix traded one bug for another. Fixed in `2ad0c2f7`.

## The regression

`modules/command_menu.js:206` only opens the menu when the text before the caret matches `/^\/([^\s/]*)$/` — a `/` token at the very start of the message. My earlier draft-preservation fix returned early for any non-empty textarea, so it moved the caret and never dispatched the `input` event the menu listens for. The "View commands" CTA did nothing at all once the user had typed anything, which is exactly the scenario the cards are visible in (they stay up while the first message is being composed).

## The fix

Prepend `/` to the draft instead of overwriting it, then park the caret at index 1:

```js
const hadSlash = textarea.value.startsWith("/")
if (!hadSlash) textarea.value = `/${textarea.value}`
const caret = hadSlash ? textarea.value.match(/^\/[^\s/]*/)[0].length : 1
textarea.setSelectionRange(caret, caret)
textarea.dispatchEvent(new Event("input", { bubbles: true }))
```

This is the same state the user would reach by pressing Home and typing `/` themselves — the draft is intact and the menu opens with the full command list. The caret position matters: putting it at the end instead would make the menu try to match the whole draft as a command name, find nothing, and `popupMenu.hide()` immediately.

The `hadSlash` branch handles a draft that already begins with a slash (`/top`): the value is left untouched and the caret goes to the end of the leading slash token, so the user's partial query pre-filters the list rather than being discarded.

## Tests

The old test that asserted "no input event" is replaced by one asserting the real contract — draft preserved, `input` dispatched, and the resulting `(value, selectionStart)` pair actually satisfies `command_menu.js`'s trigger regex with an empty query. A second test covers the partial-command case yielding query `top`. Verified the first fails against the previous early-return code.

Jest: 18 `controllers/comments` suites, 216 tests passing. Pushed with `--no-verify` per project convention.

sh1nj1:
Two Codex findings — one fixed, one I'm rebutting with a reproduction.

## Fixed: live comment arrival now clears the empty state (`072fbb69`)

Confirmed. `Comment#broadcast_create` does `broadcast_append_later_to([creative, :comments], target: "comments-list", ...)` (`broadcastable.rb:186`), and the only thing that clears the placeholder is `comments--form#removePlaceholder` (`form_controller.js:1076`), called solely from the local submit path at `form_controller.js:1061`. So when another participant posts the first public comment, the comment appends and `#no-comments` stays exactly where it is.

This did exist before the PR — the old branch rendered `<div id="no-comments">no comments</div>` and had the same gap — but a single line of stale text next to a real message is very different from a six-card discovery grid, one of whose cards advises the user to start a conversation, sitting on top of a conversation that has already started. Worth fixing here.

`feature_cards_controller` now observes its parent list in `connect()` and removes itself the first time a `.comment-item` is appended, whichever path appended it:

```js
this._listObserver = new MutationObserver((mutations) => {
  const commentArrived = mutations.some((mutation) =>
    [ ...mutation.addedNodes ].some((node) => node.nodeType === 1 && node.matches(".comment-item")),
  )
  if (commentArrived) this.element.remove()
})
this._listObserver.observe(list, { childList: true })
```

Matching on `.comment-item` rather than "any added node" keeps the cards up when something else touches the list. `disconnect()` tears the observer down. Two tests: a `.comment-item` append removes `#no-comments`, and a plain `<div>` append leaves it alone. Verified the first fails with `connect()` removed.

## Not a defect: `with_lock` does serialize on SQLite here

The premise is that `with_lock` "relies solely on a locking primitive that SQLite does not provide." That is true of the `FOR UPDATE` clause itself — the SQLite Arel visitor drops it — but it is not how the transaction gets serialized on this stack.

Rails 8.1's SQLite3 adapter begins **every** transaction as `BEGIN IMMEDIATE`:

```ruby
# activerecord-8.1.3/lib/active_record/connection_adapters/sqlite3/database_statements.rb:32
def begin_db_transaction # :nodoc:
  internal_begin_transaction(:immediate, nil)
end
```

Unconditional, not a configurable default. `IMMEDIATE` takes the write lock when the transaction opens — *before* `with_lock`'s `SELECT` — so the read-modify-write is already exclusive by the time the array is read. `config/database.yml` also sets `timeout: 5000`, so a competing request waits on `busy_timeout` instead of returning `SQLITE_BUSY`.

Reproduced on this branch, 20 threads dismissing concurrently with a 10ms sleep inserted between the read and the write to widen the race:

```
adapter=SQLite
errors=[]
expected=20 got=20 missing=[]
RESULT: NO LOST UPDATES
```

Control run, same script with `with_lock` replaced by a bare `begin`:

```
adapter=SQLite
errors=[]
expected=20 got=1 missing=["k1", "k2", ... "k20"]
RESULT: LOST UPDATES / ERRORS
```

19 of 20 writes lost without it, 0 lost with it, no `SQLITE_BUSY` in either run. The lock is doing real work on SQLite, and the harness would have caught the failure had it not been.

This is also the codebase's documented position, in `orchestration/topic_slot.rb:15-17`:

> `SELECT ... FOR UPDATE` on Postgres; the SQLite visitor drops the lock clause, where writes are serialized anyway.

`with_lock` is the convention across 11+ call sites in the engine (`AiAgentJob`, `WorkflowExecutor`, `ActionExecutor`, `ApprovalActions`, `Comment#…`, `AgentsController`). Swapping this one to a bespoke retry loop would make it the odd one out without making it safer.

**Verification:** Jest — 18 `controllers/comments` suites, 218 tests passing. Pushed with `--no-verify` per project convention.

sh1nj1:
Confirmed — and the same gap existed in the sibling state, so the fix is shared. Commit `99a7de34`.

## The bug

`comments--form#removePlaceholder` matched `#no-comments` and nothing else:

```js
const placeholder = listElement?.querySelector("#no-comments")
```

So the `#no-topic-comments` notice I added in `4ab35b09` survived the first message in an empty topic — `renderCommentHtml` appended the `.comment-item` right beside "no comments in this topic yet". `#no-search-results` had the identical problem. And on the Turbo Stream path (`Comment#broadcast_create`) neither was cleared at all, since only the discovery cards carried the mutation observer from `072fbb69`.

## The fix

Rather than adding two more `querySelector` calls and a second copy of the observer, the observer moved out of `feature_cards_controller` into a small shared `comments--placeholder` controller:

```js
export default class extends Controller {
  connect() {
    const list = this.element.parentElement
    if (!list) return
    this._listObserver = new MutationObserver((mutations) => {
      const commentArrived = mutations.some((mutation) =>
        [ ...mutation.addedNodes ].some((node) => node.nodeType === 1 && node.matches(".comment-item")),
      )
      if (commentArrived) this.element.remove()
    })
    this._listObserver.observe(list, { childList: true })
  }

  disconnect() {
    this._listObserver?.disconnect()
    this._listObserver = null
  }
}
```

All three empty-list states in `_list.html.erb` now carry `class="comments-placeholder" data-controller="comments--placeholder"` (the discovery-cards div stacks it alongside `comments--feature-cards`), and `removePlaceholder()` matches the class instead of the one id, so the local submit path clears whichever state is on screen. `feature_cards_controller` is back to just dismiss/restore/command-menu.

Matching `.comment-item` rather than "any added node" is unchanged — the placeholder stays put when something else touches the list.

## Tests

New `placeholder_controller.test.js` runs the Turbo Stream append against each of the three ids, plus a non-comment node that must leave the placeholder alone and a disconnect case proving the observer is torn down. The two observer tests are removed from `feature_cards_controller.test.js` since that behaviour no longer lives there. Verified the three id cases fail with `connect()` stubbed out.

Three Rails tests assert the rendered markup actually carries the controller, so the ERB wiring is covered and not just the JS.

**Verification:** `comments_controller_test.rb` — 63 runs, 261 assertions, 0 failures; `notices_controller_test.rb` — 3 runs, 6 assertions, 0 failures. Jest — 19 `controllers/comments` suites, 221 tests passing. Rubocop clean. Pushed with `--no-verify` per project convention.

sh1nj1:
Confirmed, and the root cause reaches past the placeholder. Commit `35cad496`.

## The bug

`comments/list_controller.js#handleStreamRender` filters live appends by topic, by dedup id, and by history mode — but never by `manualSearchQuery`. So while a search is active, a comment posted by another participant is appended into the result list whether or not it matches the query.

That is a defect on its own (the results view silently gains an unrelated message), and when the search had *no* matches it also takes out the explanation: `comments--placeholder` removes itself on any `.comment-item` arriving, so `#no-search-results` disappears and the user is left staring at one unrelated message with nothing saying why.

## The fix

Block appends at the source, in the same `if (action === 'append')` branch that already handles topic mismatch:

```js
// A search-filtered list is the result set of a query, and the match runs
// server-side over the raw content. A live append carries no verdict on
// whether it matches, so letting it in drops an unrelated message into the
// results — and, when there were none, takes the "no results" notice with
// it (comments--placeholder clears on any .comment-item arriving). Blocked
// like history mode below; the next search or reset re-queries the server.
if (this.manualSearchQuery) {
  event.preventDefault()
  return
}
```

Fixing it here rather than in the observer means the placeholder problem and the results-pollution problem both go away, and `placeholder_controller` stays a dumb "a comment arrived, I'm no longer true" rule instead of learning about search state.

Two deliberate choices:

- **Matching appends are blocked too.** Letting them through would mean reimplementing `CommentsController#index`'s filter (multi-word `LOWER(comments.content) LIKE`, splitting on whitespace, capped at `MAX_SEARCH_WORDS`) against rendered markdown in the browser — a client/server disagreement waiting to happen. A consistent snapshot beats a half-refreshed one; the next search or reset re-queries the server. This mirrors the history-mode guard directly below, which already drops live comments to keep a scrolled-back view stable.
- **Only appends are filtered.** `replace` and `remove` target a comment that is already in the result set, so suppressing them would leave stale text or a deleted message on screen. Edits and deletions stay live while search is active.

## Tests

New `list_controller_stream_search.test.js`, 6 cases: a non-matching append is blocked; a *matching* append is blocked too (pinning the deliberate choice above); appends pass through with no search active; `replace`/`remove` are never blocked; streams aimed at another target are ignored; and `resetState()` clears the filter so live comments resume. Verified the two blocking cases fail with the guard removed.

**Verification:** Jest — 20 `controllers/comments` suites, 227 tests passing. Pushed with `--no-verify` per project convention.

sh1nj1:
Confirmed. Commit `ada2f193` — same class of bug as the stream path, on the local side.

## The bug

`35cad496` blocked live appends while a search is active, but the local submit path still splices straight into the filtered list. All three submit paths funnel through `renderCommentHtml`, which calls `removePlaceholder()` unconditionally, and that now clears **any** `.comments-placeholder`. So with a search active:

- the posted message lands among the results whether or not it matches the query, and
- when the search had no matches, `#no-search-results` goes with it — leaving one message on screen with nothing explaining why it is there.

`applySearchQuery()` calls `resetState()` before setting `manualSearchQuery`, so `allNewerLoaded` is `true` during a search and the existing history-mode branch never catches this.

## The fix

Guard in `renderCommentHtml`, so all three call sites are covered (`handleSubmit`'s live-mode branch, the edit branch, and `_sendQuestionQuote`):

```js
const listCtrl = this.listController
if (!replaceExisting && listCtrl?.manualSearchQuery) {
  listCtrl.resetToLatest()
  return
}
```

`resetToLatest()` runs `resetState()` — which clears `manualSearchQuery` — then reloads from the server, so the user leaves search and sees the live thread with their new message in it. That is the same exit the history-mode branch above already takes when `!allNewerLoaded`, and it is the right outcome: sending a message means you are back in the conversation, not browsing results. Silently dropping the render instead would make the message appear to vanish.

**Edits are exempt** (`replaceExisting: true`). That comment is already in the result set, so replacing it in place is correct — and it mirrors the stream-side decision, which blocks `append` but never `replace`/`remove`, so stale text is not left on screen.

## Tests

New `form_controller_search_submit.test.js`, 4 cases: a submit during a search does not append and calls `resetToLatest()`; `#no-search-results` survives that submit; with no search active the append and placeholder removal work exactly as before; and an edit during a search still replaces in place without resetting. Verified the first two fail with the guard removed.

**Verification:** Jest — 21 `controllers/comments` suites, 231 tests passing. Pushed with `--no-verify` per project convention.

sh1nj1:
Both findings confirmed and fixed in `2ca7896e`.

## Drafts survive an argument-form command

Confirmed, and it is the deeper half of the draft problem I fixed in `2ad0c2f7`. The prepend keeps the draft only up to selection. For a schema-backed command, `command_menu.js#onSelect` runs `clearCommandText()`, which strips *only* the leading `/` token — so the draft is deliberately still in the box — then opens `CommandArgsForm`. Its submit callback then does:

```js
textarea.value = commandText
```

and clicks send. The draft that `clearCommandText` had just preserved is gone. `onCancel` was already fine (it hides and refocuses, leaving the draft alone), so the loss is specific to the submit path, and `/calendar`, `/topic`, `/work` and `/compress` all reach it.

The command is submitted on its own, so the draft genuinely cannot ride along with it — the fix is to give it back rather than to keep it in place. The menu hands it over on a bubbling `comments--form:stash-draft` event before overwriting:

```js
const draft = textarea.value
if (draft.trim()) {
  textarea.dispatchEvent(new CustomEvent('comments--form:stash-draft', {
    bubbles: true, detail: { draft }
  }))
}
```

and `comments--form` reinstates it from `.finally()`, after the send settles:

```js
_restoreStashedDraft() {
  const draft = this._stashedDraft
  this._stashedDraft = null
  if (!draft) return
  if (this.textareaTarget.value.trim().length > 0) return
  this.textareaTarget.value = draft
  this._autoResize()
  this._updateSubmitButton()
}
```

Three details worth calling out:

- **A custom event, not a controller lookup.** `command_menu.js` is a plain module with no `this.application`, and its existing `window.Stimulus` lookups are exactly what the earlier review flagged as absent in an embedding host. A bubbling event needs no global — `#new-comment-form` sits inside `#comments-popup`, which carries `data-controller="… comments--form …"` (`_comments_popup.html.erb:6,167`). There is a test pinning that bubbling contract so a future DOM reshuffle cannot silently break it.
- **Restore only into an empty box.** The success path clears the textarea before `.finally()`, but the user may have started typing again while the command was in flight, and the failure path restores review-quote text into it. Neither should be clobbered.
- **No `input` dispatch on restore.** Setting the value fires nothing, so `_autoResize()` and `_updateSubmitButton()` are called directly. Dispatching `input` instead would re-open the command menu for any draft that happens to start with `/`.

Tests: new `form_controller_stashed_draft.test.js` — the event bubbles from the textarea to the popup and is stashed; the draft is reinstated into a cleared box with resize and submit-state refreshed; text typed during the send is not clobbered; the stash is consumed so a later send cannot resurrect it; an ordinary send with nothing stashed is a no-op.

## Command card hidden for an archived Creative

Confirmed. `CommentsController#create` rejects an archived Creative at line 153, *before* the `:feedback` check, so an owner of an archived empty Creative still passes the permission filter and gets a "View commands" CTA whose every outcome is a 403. The filter now reads:

```erb
if creative.archived? || !creative.has_permission?(Current.user, :feedback)
  visible_cards = visible_cards.reject { |card| card.action&.dig(:type) == :command_menu }
end
```

Regression test added: `index hides the slash_command card for an archived creative even for the owner`. Verified it fails against the previous `unless` form.

**One thing this does not fix, deliberately:** `can_comment` is computed as `has_permission?(:feedback)` at every site that produces it (`workspace_tree_builder.rb:25`, `comments_controller.rb:267`, `creatives/index.html.erb:10`, `creatives_helper.rb:80`) and never consults `archived?`. So on an archived Creative the comment form itself is still shown and still accepts input, and submitting returns the 403 above. That is pre-existing and independent of this PR, and threading archived-ness through all four producers touches linked-creative and `effective_origin` resolution — the wrong change to bury in an empty-state PR. Happy to open an issue.

**Verification:** `comments_controller_test.rb` + `notices_controller_test.rb` — 67 runs, 272 assertions, 0 failures, 0 errors. Jest — 22 `controllers/comments` suites, 237 tests passing. Rubocop clean. Pushed with `--no-verify` per project convention.

sh1nj1:
Confirmed. Commit `b6e72ac5`.

## The bug

`resetForm()` runs only on the success path, so when a schema-backed command's request fails, the submitted command text is still sitting in the textarea when `.finally()` runs. `_restoreStashedDraft()` clears `_stashedDraft` first and *then* hits its non-empty guard, so it returns having already thrown the draft away — permanently.

The scenario is the exact one the stash exists for: type a draft, click "View commands", pick `/calendar`, fill the args form, submit, request fails. The alert says the comment failed; the draft is gone with no way back.

## The fix

`handleSend` captures the exact content it submits — after the review-quote rewrite, so it is what actually went to the server — and hands it to the restore helper:

```js
// Captured after the review-quote rewrite above, so it is the exact content
// going to the server. _restoreStashedDraft compares against it to tell a
// failure that left the command text behind from text typed mid-flight.
const submittedText = this.textareaTarget.value
```

```js
// Two boxes are safe to overwrite: an empty one (the success path runs
// resetForm) and one still holding exactly what we submitted (the failure
// path never clears it, so the command text is left sitting there). Any
// other content is text the user typed while the request was in flight, or
// a review quote the failure path restored, and must not be clobbered.
const current = this.textareaTarget.value
if (current.trim().length > 0 && current !== submittedText) return
```

Equality against `submittedText` is what distinguishes "the failure left the command here" from "the user typed something new", which is the distinction the previous guard could not make.

**On discarding the failed command text:** the draft wins. It is user-authored prose that cannot be reconstructed, and losing it is silent. The command line is machine-composed from a form the user just filled, and its loss is announced by the failure alert — re-picking the command from the menu reproduces it. The alternative, keeping the stash around to restore at some later send, is worse: the box would show the failed command while the draft reappears minutes later in an unrelated context.

The two non-restore cases are preserved: text typed during the flight, and the review-quote text `.catch` reinstates via `_reviewStore.restore()` (which never equals `submittedText`).

## Tests

New `form_controller_stashed_draft_failure.test.js` drives the real `handleSend()` against a mocked `fetch`, so the wiring is covered rather than just the helper:

- a 500 leaves the command text in the box and the draft is restored over it, with `resetForm` confirmed *not* called
- text typed while the failing request is in flight (held-open fetch promise, resolved after the edit) is not clobbered
- the success path still restores into the box `resetForm()` cleared
- a failure with nothing stashed leaves the submitted text alone, so a retry is still possible

Plus two unit cases in `form_controller_stashed_draft.test.js`: restoring over the leftover command text, and leaving a restored review quote alone. Verified both new failure-path cases fail against the previous guard.

**Verification:** Jest — 23 `controllers/comments` suites, 243 tests passing. Pushed with `--no-verify` per project convention.

sh1nj1:
Confirmed. Commit `97933b40`.

## The bug

The popup keeps a single `comments--form` controller for every conversation. Switching creatives calls `onPopupOpened` on that same instance — and `_navigateToEntry` reuses `open()`/`openForCreative()`, so `onPopupClosed()` does not even run in between (noted in `presence_controller.js:97`). `onPopupOpened` then does `resetForm()`, which empties the textarea for the newly opened creative.

So with a schema-backed command in flight for Creative A, the user switching to Creative B lands in exactly the state `_restoreStashedDraft` treats as safe: an empty box. Whichever way the request settles, A's draft is written into B's textarea, and the user's next Enter sends it to the wrong conversation.

## The fix

The stash is tagged with the conversation it came from, at stash time:

```js
handleStashDraft(event) {
  const draft = event.detail?.draft || null
  this._stashedDraft = draft ? { draft, creativeId: this.creativeId } : null
}
```

and the restore refuses to cross conversations:

```js
if (String(stashed.creativeId ?? '') !== String(this.creativeId ?? '')) return
```

Tagging at stash time rather than at send time closes a second hole in the same area: `handleSend` returns early when a send for the same creative is already in `inFlightSends`, so `.finally()` never runs and the stash outlives the send that was supposed to consume it. Without the tag, the *next* send — possibly in a different creative by then — would hand it back.

The stash is still consumed either way, so a draft that belongs to a conversation no longer on screen is dropped rather than misfiled. Restoring it into A is not an option: A's textarea is this same element, now showing B.

`String(...)` on both sides because `creativeId` arrives as a dataset string from `popup_controller` and as a number from some callers; a reopen that changes only the type is not a switch.

## Tests

New `form_controller_stashed_draft_creative_switch.test.js` drives the real `handleSend()` against a held-open `fetch`, switching creatives via the real `onPopupOpened` before the response settles:

- a switch mid-flight leaves B's box empty and consumes the stash
- reopening the *same* creative mid-flight still restores the draft
- an id that changes type across the reopen still counts as the same creative
- a stash left behind by a bailed send is not restored in another creative

Verified the first and last fail with the guard removed. `_stashedDraft` is now an object, so the existing unit tests build it through a `stash()` helper instead of assigning a bare string.

**Verification:** Jest — 24 `controllers/comments` suites, 247 tests passing; full suite 76 suites, 846 tests passing. Pushed with `--no-verify` per project convention.

PR #1487, opened 2026-08-07
fix(creatives): clear empty-state placeholder when a sub-creative is added

sh1nj1:
## Problem

On a creative page with no children, the tree shows the "No sub-creatives found."
placeholder. Adding a sub-creative inserts the new row but leaves the placeholder
visible right under it, until a full page reload.

## Root cause

The placeholder is server-rendered as a plain `<p>` inside `#creatives`
(`creatives/index.html.erb`). Rows, however, are inserted into that same container
purely client-side, and neither insertion path knew about the placeholder:

- local add — `performStart()` in `modules/creative_row_editor.js`
- remote / MCP add — `handleCreated()` in `lib/turbo_stream_actions.js`

Nothing identified the placeholder element, so nothing could remove it.

## Fix

- Wrap the placeholder in a `data-creatives-empty-state` marker element.
- New shared module `modules/creative_tree_empty_state.js`:
  - `hideTreeEmptyState()` — hides the placeholder when a row lands.
  - `restoreTreeEmptyState()` — shows it again once the last row is gone; no-op
    while any row remains.
  - The node is toggled (`hidden` + `display:none`) rather than removed and
    re-created from cached markup: no HTML re-parse (CodeQL `js/xss-through-dom`),
    and the "request permission" variant keeps its server-rendered `button_to`
    CSRF token.
- Hooked into both creation paths, and into both removal choke points
  (`removeTreeElement()` for editor-side removals — abandoned new row, delete —
  and `handleDestroyed()` for broadcasts), so the placeholder comes back instead
  of leaving an empty tree with no message.

## Tests

- `modules/__tests__/creative_tree_empty_state.test.js` — new module (hide,
  restore, guards for missing container / missing placeholder / no duplicates).
- `lib/__tests__/turbo_stream_actions.test.js` — broadcast `created` hides the
  placeholder, `destroyed` restores it for the last row and leaves it hidden
  while siblings remain.
- `modules/__tests__/creative_tree_dom.test.js` — `removeTreeElement` restore
  behavior.
- `test/system/creative_inline_edit_test.rb` — end-to-end: empty parent page,
  add button, placeholder gone.

Verified red/green: the system test fails against the pre-fix bundle
(placeholder still rendered next to the new row) and passes after the change.
All 770 engine JS tests and all 9 tests in `creative_inline_edit_test.rb` pass
locally; the full CI `test` job passed on the first commit.

### DIFF
--- a/engines/collavre/app/javascript/controllers/creatives/__tests__/select_mode_controller.test.js
+++ b/engines/collavre/app/javascript/controllers/creatives/__tests__/select_mode_controller.test.js
@@ -1,0 +1,139 @@
+/**
+ * @jest-environment jsdom
+ */
+import { jest } from '@jest/globals'
+
+// Batch delete is DOM-repaired entirely client-side: the destroy broadcast
+// excludes the initiating user, so nothing comes back to fix this window.
+const csrfFetch = jest.fn(() => Promise.resolve({ ok: true }))
+const confirmDialog = jest.fn(() => Promise.resolve(true))
+
+jest.unstable_mockModule('../../../lib/api/csrf_fetch', () => ({
+  default: csrfFetch,
+  updateCsrfTokenFromResponse: jest.fn(),
+  refreshCsrfToken: jest.fn(),
+}))
+jest.unstable_mockModule('../../../lib/utils/confirm_dialog', () => ({
+  default: confirmDialog,
+  confirmDialog,
+}))
+
+const { Application } = await import('@hotwired/stimulus')
+const SelectModeController = (await import('../select_mode_controller')).default
+
+const EMPTY_HTML = '<div data-creatives-empty-state=""><p>No sub-creatives found.</p></div>'
+
+const flush = () => new Promise((resolve) => setTimeout(resolve, 0))
+
+function rowMarkup(id) {
+  return `
+    <creative-tree-row creative-id="${id}">
+      <div class="creative-tree" id="creative-${id}" data-id="${id}" data-level="1">
+        <div class="creative-row" data-creatives--select-mode-target="row">
+          <input type="checkbox" class="select-creative-checkbox" value="${id}"
+                 data-creatives--select-mode-target="checkbox" />
+        </div>
+      </div>
+    </creative-tree-row>
+  `
+}
+
+// Post-client-render state: tree_controller wiped #creatives and rendered the rows
+// into it, so the placeholder only survives in the out-of-container template. Rows
+// are injected after the controller connects, as they are in the real app — the
+// server always ships #creatives empty and the tree fetch fills it in.
+async function mount(rowIds) {
+  document.body.innerHTML = `
+    <template id="creatives-empty-state-template">${EMPTY_HTML}</template>
+    <div data-controller="creatives--select-mode">
+      <button id="delete-selection-btn" data-confirm="Are you sure?"
+              data-creatives--select-mode-target="deleteButton"></button>
+      <div id="creatives"></div>
+    </div>
+  `
+
+  const application = Application.start()
+  application.register('creatives--select-mode', SelectModeController)
+  await flush()
+
+  container().innerHTML = rowIds.map(rowMarkup).join('')
+  await flush()
+  return application
+}
+
+function controllerFor(application) {
+  return application.getControllerForElementAndIdentifier(
+    document.querySelector('[data-controller="creatives--select-mode"]'),
+    'creatives--select-mode'
+  )
+}
+
+function container() {
+  return document.getElementById('creatives')
+}
+
+function placeholder() {
+  return container().querySelector('[data-creatives-empty-state]')
+}
+
+async function deleteSelected(application, ids) {
+  ids.forEach((id) => {
+    container().querySelector(`.select-creative-checkbox[value="${id}"]`).checked = true
+  })
+  await controllerFor(application).deleteSelected(new Event('click'))
+  await flush()
+}
+
+let application
+
+afterEach(() => {
+  application?.stop()
+  application = undefined
+  jest.clearAllMocks()
+  document.body.innerHTML = ''
+})
+
+test('deleting every selected row restores the empty-state placeholder', async () => {
+  application = await mount(['7', '8'])
+
+  await deleteSelected(application, ['7', '8'])
+
+  expect(csrfFetch).toHaveBeenCalledTimes(2)
+  expect(csrfFetch.mock.calls.map((call) => call[0])).toEqual([
+    '/creatives/7?delete_with_children=false',
+    '/creatives/8?delete_with_children=false',
+  ])
+  expect(placeholder()).not.toBeNull()
+  expect(placeholder().hidden).toBe(false)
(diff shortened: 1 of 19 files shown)


sh1nj1:
Both review comments were against `c55ee7ff`. Status after `f140fbd5`:

**CodeQL — `js/xss-through-dom` (alert #60)**

Fixed in `71ec0987`: the `insertAdjacentHTML` restore path was dropped in favour of toggling the existing node. `f140fbd5` goes further and removes the remaining HTML sink in this area — `showEmptyState()` no longer does `this.element.innerHTML = this.emptyHtmlValue`, and the `creatives--tree-empty-html-value` attribute is gone.

**Codex P2 — "Cache empty markup for initially populated trees"**

The stated mechanism is not accurate: `index.html.erb` never sees a populated tree. `CreativesController#index` sets `@creatives = []` unconditionally on the HTML branch ("CSR will fetch via JSON"), so `(@creatives || []).blank?` is always true and the placeholder is always built.

The conclusion is right, though, for a different reason. The tree is client-rendered, and `tree_controller` wipes the container on every load (`showLoadingIndicator()` → `this.element.innerHTML = ''`, then `renderCreativeTree`). On a tree that has rows at load time the placeholder is therefore gone by the time the rows land, and the restoration path added in this PR had nothing to un-hide when the last row was deleted.

`f140fbd5` fixes that by rendering the placeholder into a `<template id="creatives-empty-state-template">` **outside** `#creatives`, which survives the wipe. `restoreTreeEmptyState()` un-hides an existing placeholder when there is one, and otherwise clones the template. Cloning `template.content` is not an HTML sink (the browser parsed the server-escaped markup at page load), and unlike re-parsing a cached string it preserves the `button_to` CSRF token in the request-permission variant.

Coverage added: 5 jsdom tests for the clone path (including template-not-mutated-by-hide), 3 controller tests for the template placement, and a system test asserting the template survives a populated CSR render.

sh1nj1:
Both P2 comments are correct — confirmed and fixed in `7af1e086`, plus a third instance of the same class and a latent `ReferenceError` the new tests exposed.

**"Restore the empty state after archiving the last row"** — confirmed. `Creative#archive!` (`creative.rb:303`) is a chain of `update_all` calls, so no callback runs and no destroy broadcast is emitted; the archive click handler removed the row with a bare `row.remove()`. It now goes through `removeTreeElement()`.

**"Restore the placeholder after batch deletion"** — confirmed on both counts. `deleteSelected` removed `document.getElementById('creative-<id>')`, which is the inner `.creative-tree` div (`broadcast_node_payload` sets `dom_id: "creative-#{id}"`), leaving an empty `<creative-tree-row>` wrapper that `restoreTreeEmptyState()` reads as "still populated". And `broadcast_creative_destroyed` passes `current_user_id: broadcast_excludable_user_id`, so the initiating window never gets its own destroy event. Also routed through `removeTreeElement()`, which drops the wrapper.

**Third instance, same class** — `removeDroppedCreative()` in `creatives/drag_drop/event_handlers.js`: dragging a row into another window removes it from the source window's tree, and a move is not a destroy, so nothing repairs the placeholder there either. Now restores it.

**Pre-existing `ReferenceError` surfaced by the new archive test** — the archive handler ended with `closeEditor()`, and no function by that name exists anywhere in the codebase (`grep -rn closeEditor` returns exactly that one call site). Archiving therefore threw after removing the row, leaving the inline editor open over the archived creative. Replaced with the module's real `hideCurrent()` — the same function the close button binds to — moved ahead of the DOM mutations so the editor is not closed against a detached tree.

**Coverage** — 11 new jsdom tests: 4 for batch delete (including "the wrapper goes with the row"), 4 for archive (last row, one-of-many, declined confirmation, and the unarchive/reload branch), 3 for the cross-window drop. Each fix was verified to fail its test when reverted. The inline-editor DOM fixture is now shared via `modules/__tests__/support/inline_editor_dom.js` instead of duplicated.

Local: 834 JS tests / 72 suites green; patch lines fully covered.

sh1nj1:
Confirmed and fixed in `2aee8c18`.

`hideCurrent()` flushes a pending edit via `saveForm()`, and `saveForm()`'s `.catch` explicitly rethrows after setting the `error` save status (`creative_row_editor.js:879-882`). A dropped connection therefore rejected the promise the archive handler was chained onto, so `removeTreeElement(row)` never ran — and since `Creative#archive!` is an `update_all` there is no destroy broadcast to repair the view. The row stayed on screen pointing at a creative that is archived on the server until a reload.

The fix applies the view change regardless of how the flush settles: the close is now `Promise.resolve().then(hideCurrent).catch(log).then(applyToView)`. Archiving first and flushing after is kept deliberately — reordering would mean a flaky save could cancel an archive the user already confirmed. `hideCurrent()` sets `template.style.display = 'none'` and clears `currentTree` synchronously, before the save promise is awaited, so the editor is genuinely closed on the rejecting path; only the removal was missing. `waitForUploads()` never rejects, so the save is the only rejection source here.

Two jsdom tests added, both driving a real editor session (`creative-edit-click` on the row, then the captured `onChange` to leave a pending save): one asserts the pending edit is flushed on the happy path, one asserts the row is dropped and the placeholder restored when `creativesApi.save` rejects. The failure test was verified to fail when the `.catch` is removed.

Local: 836 JS tests / 72 suites green.

sh1nj1:
Confirmed and fixed in `e8055d59`.

The Chats feed is genuinely paginated (`IndexQuery#comment_chats_page`, `COMMENT_CHATS_PER_PAGE`), and `_setupPagination()` leaves a sentinel in `#creatives` whenever `has_more` is true. Batch-deleting every row that is currently rendered therefore empties the container without emptying the result set — `restoreTreeEmptyState()` saw no `creative-tree-row` and put the placeholder back. Worse, the now-empty container puts the sentinel straight into the viewport, so the IntersectionObserver fires immediately and `_loadMore()` appends page 2 through `appendCreativeNodes()`, which has no empty-state hook. "No creatives found." ends up sitting above the rows it just contradicted.

The fix marks the container while pages are outstanding. `_setupPagination()` sets `data-creatives-pagination-pending` alongside the sentinel and `_teardownPagination()` removes it, and `restoreTreeEmptyState()` returns early when that attribute is present — an empty container is not an empty feed, it is a feed between pages. Two supporting changes:

- `_loadMore()` calls `hideTreeEmptyState()` before appending, so a placeholder that reached the DOM by some other route still cannot survive an append.
- When the final page lands (`has_more: false`), the teardown is followed by an explicit `restoreTreeEmptyState()`. Nothing is queued any more, so an empty container at that point genuinely is an empty feed — this covers the case where the rows that were on screen when page N was requested got deleted while it was in flight.

The load-more error path deliberately leaves the marker set: the sentinel stays in place for a retry, so the pages are still pending, not absent.

Coverage: 4 controller tests (marker set with the sentinel; placeholder suppressed after every rendered row is removed; placeholder restored when the last page arrives empty; marker cleared when `load()` tears pagination down) and 3 module tests (clone path suppressed, existing placeholder stays hidden, restore resumes once the marker clears). All 7 were verified to fail when the guard and the append-side hide are reverted.

Local: 843 JS tests / 72 suites green.

PR #827, opened 2026-02-16
feat: Add landing page for unauthenticated users

sh1nj1:
## Summary

Add a marketing landing page at `/` for non-authenticated users. Authenticated users are redirected to the creatives index.

## Changes

- **PagesController#landing** — serves the landing page, redirects authenticated users to `creatives#index`
- **Landing layout** (`layouts/landing.html.erb`) — standalone layout with no app navigation, loads only design tokens + landing CSS
- **Landing view** — hero section, pain points, features, AI agent showcase, dogfooding section, and final CTA
- **Landing CSS** — responsive styles using Collavre design tokens (no external frameworks)
- **i18n** — full en + ko locale files for all landing page text
- **Route change** — root now points to `pages#landing` instead of `collavre/creatives#index`

## Design

- Uses `--color-brand` (oklch green) as accent color
- Inter font via `--font-sans`
- Sticky frosted-glass nav bar
- Responsive: 3-column grid on desktop, single column on mobile
- App screenshot placeholder with mock UI elements
- AI agent flow visualization

## Testing

- Visit `/` as unauthenticated user → landing page
- Visit `/` as authenticated user → redirect to creatives index
- Rubocop passes with no offenses

PR #1489, opened 2026-08-07
feat: add feature guides and show them in empty topics

sh1nj1:
## Summary

Phase 2 of the empty-chat feature discovery work started in #1486.

Phase 1 shipped the six discovery cards with their "learn more" link suppressed because a `guide_enabled: true` flag that pointed at a missing route would crash. This PR adds the pages, turns the links on, and displays the same feature guidance when the selected topic has no messages.

## What's here

**Routes and controller**
- `GET /features` — public hub listing every card that has a guide
- `GET /features/:key` — the long-form guide for one card
- Both are readable signed out and use the landing layout + stylesheet, so a guide reached from a shared link reads as the same product surface as the landing page beside it.

**Registry**
- `FeatureCard` gains `guide: true` to opt into the engine-provided page, alongside the existing `guide_url:` for vendor engines hosting their own docs.
- A card carrying its own `guide_url` still renders the link, but gets no engine page: `#show` 404s for it and the hub omits it. This keeps the route from serving an empty shell of headings for a card documented elsewhere.

**Empty topics**
- An empty topic now renders the feature discovery cards instead of the standalone "No messages in this topic yet" notice.
- The obsolete topic-only placeholder, translations, and extra comment existence query are removed.
- Search results with no matches keep their dedicated empty state.

**Copy**
- All guide content lives in `config/locales/features.{en,ko}.yml` under `collavre.features.pages.<key>` — title, tagline, meta description, sections, and tips, in both languages.
- Content is written against the actual implementation: the `@name:` mention pattern from `MentionParser`, the real command set from `CommandMenuService` (`/calendar`, `/topic`, `/work`, `/compress`, `/creative` plus MCP tools), the DONE/CONTINUE/BLOCKED protocol from `trigger_loop`, and context inheritance.
- The controller normalizes `sections` and `tips` before render, because I18n returns a `String` for a missing key where the view walks an `Array`. A page can ship before its sections exist without 500ing.

**Layout**
- `layouts/collavre/landing` now reads `:page_title` / `:page_description` via `content_for`, falling back to the landing copy. Guides get their own `<title>`, `description`, and `og:` tags instead of inheriting the landing page's.

**Landing**
- The features section gains a link through to the hub.

## Testing

- `mise x -- bin/rails test engines/collavre/test/controllers/comments_controller_test.rb` — 66 runs, 292 assertions, 0 failures
- `npm test -- --runInBand engines/collavre/app/javascript/controllers/comments/__tests__/placeholder_controller.test.js engines/collavre/app/javascript/controllers/comments/__tests__/feature_cards_controller.test.js` — 12 tests, 0 failures
- `mise x -- bin/rubocop engines/collavre/app/controllers/collavre/comments_controller.rb engines/collavre/test/controllers/comments_controller_test.rb` — no offenses
- The public landing route returned HTTP 200 from the running preview server after the change.

The existing guide coverage also verifies the hub and all guide pages in English and Korean, metadata, missing-copy handling, vendor guide URLs, feature-card links, route mount prefixes, registry behavior, and landing navigation.

## Still open (not in this PR)

- Phase 3: state-aware cards (trigger on/off, etc.)
- Engine-wide request-aware base path for the JS call sites that hardcode absolute URLs
- `can_comment` does not reflect `archived?` at its production call sites


sh1nj1:
## Codex review response

**[P2] Route constraint vs registry keys — fixed** in `f5d9ae22`.

Real gap. The registry accepted any symbolizable key while the route only routes `[a-z0-9_]+`, so an extension registering `:"vendor-card"` with `guide: true` would raise `UrlGenerationError` from `feature_path` the first time the public hub or any user's empty chat rendered its link. An extension typo becoming a 500 for end users is the wrong failure mode.

Fixed by rejecting the key at registration instead of at render, with a message pointing at `:guide_url` as the escape hatch, so it fails at boot where the extension author sees it. `routes.rb` now reuses `FeatureCard::GUIDE_KEY_FORMAT` verbatim and a test asserts the route's `path.requirements[:key]` is that same object, so the two cannot drift.

One wrinkle worth recording: Rails rejects anchor characters in a routing requirement (`Regexp anchor characters are not allowed in routing requirements`), so the shared constant is unanchored for the route and `GUIDE_KEY_FORMAT_ANCHORED` is derived from its `.source` for validation. Single source of truth, correct semantics at both ends.

---

**[P1] Mount prefix in guide links — not changing.** The stated mechanism is wrong.

The claim is that `collavre` "returns `Collavre::Engine.routes.url_helpers`, which is not request-aware." It does not. In a view or controller, the mounted helper returns an `ActionDispatch::Routing::RoutesProxy` that *wraps* those helpers, and the proxy is precisely the request-aware layer. From `actionpack-8.1.3/lib/action_dispatch/routing/routes_proxy.rb:33-45`:

```ruby
def method_missing(method, *args)
  if @helpers.respond_to?(method)
    options = args.extract_options!
    options = url_options.merge((options || {}).symbolize_keys)

    if @script_namer
      options[:script_name] = merge_script_names(
        options[:script_name],
        @script_namer.call(options)
      )
    end

    args << options
    @helpers.public_send(method, *args)
```

Every generated path merges `url_options` from the calling scope (which carries `script_name` off the request) and then applies the engine's `@script_namer`. That is the mount-prefix mechanism, and `collavre.feature_path` goes through it.

Empirically, the new links behave identically to what already ships. Rendering `/features` with `SCRIPT_NAME` set produced `/features/mention_agent` alongside `/landing` and `/users/new` — and those last two are pre-existing `collavre.landing_path` / `collavre.new_user_path` calls from before this PR.

That is the substance of the objection: this PR adds 6 call sites to a pattern the engine's views already use in **87** places, including `collavre.creative_creative_shares_path` in the very partial being flagged. Whatever is or isn't true about subpath mounting, it is uniformly true across all 93. Special-casing these 6 would fake a guarantee the app does not have — which is the same conclusion reached on the equivalent finding in #1486, where it was split out as its own follow-up.

Subpath mounting stays where it was left: a separate piece of work that either makes the whole engine request-aware and documents subpath mounting as supported, or does not. It does not belong in a PR about guide pages.

sh1nj1:
## Codex review response — both accepted

### [P2] Feature-guide breadcrumbs hidden — fixed in `17d3bcfc`

Correct, and worse than "not visible enough": the breadcrumbs never rendered to a user at all.

`.landing-page nav { display: none }` is (0,1,1); `.feature-guide-breadcrumb { display: flex }` is (0,1,0). The old rule wins regardless of order.

Confirmed in a browser rather than by reading specificity — with the rule still in place, the new system test fails on both guide pages:

```
FeatureGuideTest#test_the_hub_breadcrumb_is_visible
  expected to find visible css "nav.feature-guide-breadcrumb" but there were no matches.
  Also found "", which matched the selector but not all filters.
FeatureGuideTest#test_a_guide_breadcrumb_is_visible_and_walks_back_to_the_hub
  (same)
3 runs, 2 failures
```

After the fix: `3 runs, 7 assertions, 0 failures`.

**On the fix chosen.** Rather than out-specifying the old rule, I removed it. It could not do what it was written for:

- `landing.css` is linked from exactly one place — `layouts/collavre/landing.html.erb`.
- That layout's body is `<main><%= yield %></main>`. It renders no navigation.
- App navigation is `collavre/shared/_navigation.html.erb`, rendered only by the host's `application.html.erb`, whose body class is never `landing-page` and which never loads this stylesheet.

So the selector could only ever match a `<nav>` authored inside a landing-layout view. The breadcrumb was the first one to exist, which is why this surfaced now. It was added with the landing page in #827 and has been inert since. Out-specifying it would have left the trap armed for the next `<nav>` on a landing page.

A response-body assertion can't catch this class of bug — the markup is present either way — so the regression guard is a system test, which CI runs via `rake test:system`.

### [P1] Guide-card arrow outside i18n — fixed in `0dda3d77`

Fair, and it was inconsistent with this PR's own locale files. Every other arrow-bearing label already ships the glyph inside the string:

- `landing.features.explore: "Explore all features →"`
- `features.nav.all_arrow: "← All features"` (added in this PR, used by `show.html.erb`)

The hub card was the sole place concatenating one in the view. Replaced with a dedicated `features.index.card_more` key — `"Learn more →"` / `"자세히 보기 →"` — so the whole call to action, glyph included, is reorderable by a translator. Asserted in the controller test for both locales.

---

The earlier [P1] on the mount proxy is unchanged from [my previous response](https://github.com/sh1nj1/plan42/pull/1489#issuecomment-5216064596): these links behave identically to the 87 existing engine-helper call sites, so fixing only the new ones would fake a guarantee the app does not make. Tracked separately.

Local: `86 runs, 429 assertions, 0 failures` (controller + registry + comments), `3 runs, 0 failures` (system), rubocop clean.

PR #1491, opened 2026-08-07
perf: scope pre-push checks to changed files

sh1nj1:
## Summary

- derive changed files from the commits included in each push
- run RuboCop only for changed Ruby, Rake, gemspec, Gemfile, and Rakefile paths
- map changed host and engine source files to their existing related test files
- support targeted `rake test` runs through the `TEST` environment variable
- keep the configured `.githooks/pre-push` entry point in sync through a shared implementation

## Why

The pre-push hook ran RuboCop across the entire repository and executed the full non-system test suite on every push. Those checks are already covered in CI and made local pushes unnecessarily slow.

## Validation

- `./bin/rubocop -a` (1,142 files)
- `RAILS_ENV=test rake test` (3,652 runs, 11,697 assertions)
- `RAILS_ENV=test rake test:system` (60 runs across engine system suites, 0 failures)
- simulated a new-branch pre-push update; it selected 3 RuboCop targets and 1 related test file


sh1nj1:
Closing in favor of a smaller follow-up that only scopes RuboCop to changed files and temporarily disables pre-push tests because CI runs the full suite.

PR #1495, opened 2026-08-07
perf: lazily load workspace tree branches

sh1nj1:
## Summary

- replace the eager six-level workspace-tree payload with a lightweight `expand[]` contract
- batch every visible tree level and preserve linked-shell permission behavior with cycle protection
- keep expanded branches across invalidations while restoring scroll, focus, request ownership, and failed deep-link retries
- cap expanded branch IDs at 100 on both the client and server

## Why

The persistent workspace tree built and rendered every branch up to the display-level limit on every initial load and invalidation. Its depth-first recursion also invoked batching once per node, so collapsed branches still caused hundreds of database queries and unnecessary JSON/DOM work.

## Impact

The initial request now returns root branches only, and subsequent requests return the currently visible tree in one response. A synthetic 120-node tree produced 3 nodes / 15 queries / 441 bytes initially and 9 nodes / 16 queries / 1.2 KB for one expanded path. Paths are no longer truncated at six levels.

## Validation

- Jest: 959 tests
- Rails unit/integration: 3,695 runs / 12,203 assertions
- Rails system: 67 runs / 388 assertions (2 existing skips)
- Rubocop: 1,145 files / 0 offenses
- asset build
- focused workspace system: 10 runs / 72 assertions
- Vrex review: no blocker or major findings

### DIFF
--- a/engines/collavre/app/controllers/collavre/creatives_controller.rb
+++ b/engines/collavre/app/controllers/collavre/creatives_controller.rb
@@ -1,5 +1,7 @@
 module Collavre
   class CreativesController < ApplicationController
+    WORKSPACE_TREE_EXPANSION_LIMIT = 100
+
     include Collavre::Concerns::SlideViewable
     include Collavre::Concerns::Exportable
     include Collavre::Concerns::TreeManageable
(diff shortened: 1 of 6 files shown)


sh1nj1:
CI note: the `test_js` job passed Jest and coverage, then failed only at `npm audit --production` because `main` currently pins DOMPurify 3.4.12, which is covered by GHSA-55q2-fjhq-7xh7 (`<=3.4.12`). The latest `main` CI run fails at the same audit step, while its Rails test job passes. This dependency advisory is unrelated to the workspace-tree diff and should be handled as a separate security dependency update.

PR #1632, opened 2026-09-08
feat(dnd): connect workspace and creative trees through shared input

sh1nj1:
Dragging between the workspace and creative trees uses the T1 public data/session/hit-test contract and the T2 movement command. The workspace adapter refreshes server-confirmed placement while preserving expansion, focus and scroll; the creative adapter retains DOM recovery.

Includes delegated native inputs, stored drop intent, narrow-row hysteresis, 600ms lazy branch expansion with retry after failed loads, stale-response protection, known descendant rejection, complete cross-window bundle synchronization, and translated partial-link feedback. Touch, comment-source migration, bundles and movement menus are separate follow-up PRs.

Validation: the latest T3 hover fix passed 2,027 Jest tests, 15 related system tests / 106 assertions, build, RuboCop and complexity. The integrated candidate abccc65a passed 154 Jest suites / 2,172 tests and build. Its prior tree dd6d6d57 passed all 4,832 Rails tests / 16,867 assertions, 13 focused UI/helper regressions, RuboCop and complexity on Mac. Independent review checked all 18 prior review findings and passed 205 focused tests; all findings are addressed. Latest CI must pass before merge.

Targets feature/dnd-expansion after T1 and T2. Only a local single right-tree move is optimistic; workspace and bundle placements refresh after server success.

### DIFF
--- a/engines/collavre/app/assets/stylesheets/collavre/workspace.css
+++ b/engines/collavre/app/assets/stylesheets/collavre/workspace.css
@@ -103,8 +103,26 @@ body.creative-workspace .indent3 {
 
 .creative-workspace-tree-row {
     display: flex;
+    position: relative;
     align-items: center;
     min-width: 0;
+}
+
+.creative-workspace-tree-row.is-dragging {
+    opacity: 0.55;
+}
+
+.creative-workspace-tree-row.drag-over-top {
+    box-shadow: inset 0 2px 0 var(--color-active);
+}
+
+.creative-workspace-tree-row.drag-over-bottom {
+    box-shadow: inset 0 -2px 0 var(--color-active);
+}
+
+.creative-workspace-tree-row.drag-over-child {
+    background: color-mix(in srgb, var(--color-active) 12%, transparent);
+    border-radius: var(--radius-2);
 }
 
 .creative-workspace-tree-branch-toggle,
--- a/engines/collavre/app/javascript/components/creative_tree_row.js
+++ b/engines/collavre/app/javascript/components/creative_tree_row.js
@@ -238,9 +238,6 @@ class CreativeTreeRow extends LitElement {
 
     const dragEnabled = !this.selectMode || this.canWrite;
     const draggableAttr = dragEnabled ? "true" : nothing;
-    const dragActions = dragEnabled
-      ? "dragstart->creatives--drag-drop#start dragover->creatives--drag-drop#over drop->creatives--drag-drop#drop dragleave->creatives--drag-drop#leave"
-      : nothing;
 
     return html`
       <div
@@ -250,7 +247,6 @@ class CreativeTreeRow extends LitElement {
         data-parent-id=${this.parentId ?? ""}
         data-level=${this.level ?? nothing}
         draggable=${draggableAttr}
-        data-action=${dragActions}
       >
         <div class="creative-row level-${this.level}" data-creatives--select-mode-target="row">
           <div class="creative-row-start">
--- a/engines/collavre/app/javascript/controllers/__tests__/workspace_tree_controller.test.js
+++ b/engines/collavre/app/javascript/controllers/__tests__/workspace_tree_controller.test.js
@@ -12,6 +12,8 @@ describe('WorkspaceTreeController', () => {
   let fetchMock
   let preventNavigation
 
+
+
   beforeEach(async () => {
     window.localStorage.clear()
     fetchMock = jest.fn().mockImplementation((url) => Promise.resolve(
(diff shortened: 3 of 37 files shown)


sh1nj1:
## Review follow-up: coverage + 3 Codex findings

Pushed as `c5f3ef5` (coverage) and `2b19dd7` (fixes).

### Codecov: 91.20% → 61 lines missing

60 of the 61 are now covered. Verified locally by intersecting the `main...HEAD` added-line set with an istanbul run of the full suite:

| File | Patch lines missing |
|---|---|
| `lib/dnd/registry.js` | 15 → **0** |
| `creatives/drag_drop/event_handlers.js` | 34 → **1** |
| `creatives/drag_drop/workspace_tree_adapter.js` | 5 → **0** |
| `controllers/workspace_tree_controller.js` | 6 → **0** |
| `controllers/creatives/tree_controller.js` | 1 → **0** |

The one remaining line is `getRowByCreativeId`'s `typeof document === 'undefined' || !creativeId` guard. It is unchanged code that the diff only *relocated*, and every caller already guards `creativeId` before reaching it, so there is no honest way to execute it from a public entry point. Left as-is rather than deleting a defensive guard or writing a test that reaches into module internals.

### P2 — Load lazy children before expanding on hover ✅ fixed

Confirmed. `tree_renderer.js` renders a collapsed branch with `data-loaded="false"` and a `data-load-url`, so `setExpanded` only made an empty container visible — the hover expansion advertised drop targets it never produced. Hover expansion now fetches the branch through `creativesApi.loadChildren` and renders it before revealing the container, matching `expansion_controller#ensureLoaded`. A failed fetch leaves the row collapsed and logs.

### P2 — Do not broadcast link drops as source moves ✅ fixed

Confirmed, and it is the worst of the three: the source window's `syncSourceWindowDrop` takes the removal path when the detail carries no `targetTreeId`, so a cross-window Shift-drag deleted the very row the link was made from. `notifyMoveCompletion` now takes `mode` and only emits the storage signal for `move`. The same-window completion event still fires for both, since the trees have to refresh either way.

### P2 — Surface partial link failures ✅ fixed

Confirmed as a silent failure, with one correction to the report: the DOM is not left wrong — `runMoveWithDomRecovery` already reverts on any non-`SUCCESS` status, and the completion dispatch has to stay so the rows that *did* move get refreshed. The actual gap was that the user was never told. Added `drag_drop/move_feedback.js`, called by both trees, which logs the failed ids and surfaces the server's own (already localized) rejection message. It stays silent when the server sends no message rather than inventing an untranslated client-side string.

### Verification

- Jest: 142 suites, 1906 tests passing
- System tests: `workspace_tree_drag_drop_system_test.rb` — 3 runs, 15 assertions, 0 failures
- `npm run build` clean

Still pending, unchanged: this branch carries the earlier T1 commit and needs a rebase onto the final #1630 / #1629 SHAs before merge.

sh1nj1:
## Review follow-up 2: 4 Codex findings on `2b19dd7f2`

All four confirmed and fixed in `66969a2` — no rebuttals this round.

### P2 — Permit copy effects for right-tree drags ✅ fixed

Confirmed, and it is the one that breaks a whole direction of the feature. `event_handlers.js#handleDragStart` set `effectAllowed = 'move'` while `workspace_tree_adapter.js` answers a shift-drag with `dropEffect = 'copy'`. The browser negotiates that pair down to no drag operation, so right→left link never delivered a drop event at all — no-drop cursor, nothing happens.

Fixed on the source side: the central tree now advertises `copyMove`. Its own `dragover` also reports `event.shiftKey ? 'copy' : 'move'` instead of always `'move'`, so the cursor stops claiming a move during a link.

### P2 — Invalidate hover expansion responses on tree reload ✅ fixed

Confirmed. `load()` did not touch `dragExpandAbortController`, and the expansion continuation had no generation guard, so a response generated before the move could land after the authoritative reload — splicing stale children in and replacing `nodesData` with the old placement.

Two changes: `load()` aborts the pending expansion up front, and the continuation captures `loadRequestId` and bails if a newer load has since claimed the tree. The abort alone was not enough — a response already read when the abort fires resolves normally.

### P2 — Clear aborted branch IDs before skipping expansion ✅ fixed

Confirmed. Hover A → B aborted A's request but left A in `expandedCreativeIds`, and B's response committed a set containing both while only rendering B. Coming back to A then hit `has(A)` and returned, so A could never be expanded again for the rest of the drag.

The shortcut now reads the rendered row (`item.dataset.expanded`), which is the only state that reflects what was actually drawn, and an aborted id that was never committed is removed from the set so a later `load()` does not replay it. Added `findWorkspaceItem`, which `renderExpandedBranch` now shares.

### P2 — Load children when restoring an expanded row ✅ fixed

Confirmed — this is the same defect as the previous round's hover finding, one layer up, and the earlier fix did not reach it. A hover expansion is not persisted, so the post-drop view-preserving reload renders that branch with `data-loaded="false"`, and `restoreExpansion` called `setExpanded` straight onto the empty container.

The lazy loader added last round moved out of `event_handlers.js` into `creatives/branch_expansion.js`, and `tree_view_state.js` restoration now goes through it. Both trees share one path to "expand a branch, fetching it first if it was never loaded", matching `expansion_controller#ensureLoaded`.

### Verification

- Jest: 142 suites, **1913 tests** passing
- Patch coverage: **100%** of added lines and branches across all four touched files, measured by intersecting the diff against `2b19dd7f2` with an istanbul run of the full suite
- System tests, after `npm run build` (they read built assets, not source):
  - `workspace_tree_drag_drop_system_test.rb` — 3 runs, 15 assertions, 0 failures
  - `drag_drop_system_test.rb` + `workspace_tree_drawer_test.rb` — 10 runs, 71 assertions, 0 failures (the effect-negotiation change touches the shared source path, so the neighbouring suites were re-run too)

Still pending, unchanged: this branch needs a rebase onto the final #1630 / #1629 SHAs before merge.

sh1nj1:
## Review follow-up 3: 4 findings on `66969a2`

All four findings were confirmed and fixed in `02aab342e`.

1. **Localized partial-failure fallback**: move results now distinguish server-payload messages from diagnostic exception text. Empty 403/422 responses and transport failures use EN/KO copy rendered into both tree controllers, so strings such as `HTTP 403: Forbidden` are never shown as UI fallback.
2. **Ordered nested view restoration**: expansion restoration is asynchronous and processes rows in DOM ancestry order, awaiting each lazy load before locating descendants or restoring focus.
3. **Single in-flight hover expansion**: a per-row pending marker remains until the lazy-load promise settles, preventing repeated 600 ms timers and duplicate requests on slow connections.
4. **Empty lazy branches collapse cleanly**: after rendering, the helper checks for actual child rows. An empty response clears `has-children`, collapses the row, and leaves the container loaded and hidden.

Verification:

- Jest: 143 suites, 1,918 tests passed
- Focused coverage: 8 suites, 145 tests passed; added JavaScript lines covered
- Rails controller: 47 runs, 269 assertions passed
- Three-direction DnD system test: 3 runs, 15 assertions passed
- Adjacent DnD/drawer system tests: 10 runs, 71 assertions passed
- `npm run build` passed
- `bin/complexity_check` passed under Ruby 3.4.4
- Targeted RuboCop passed

@codex review

sh1nj1:
## Review follow-up 4: empty workspace hover branch

Confirmed and fixed in `7b2010621`.

When the refreshed workspace payload returns no children, the hover expansion now removes the stale child list, clears `hasChildren` and `expanded`, replaces the branch toggle with a spacer, and excludes the creative from both pending and committed expansion state. This matches the central-tree lazy expansion behavior and lets a future authoritative reload recreate the affordance if children reappear.

Regression coverage includes empty root and nested branches, with and without an existing toggle.

Verification:

- Jest: 143 suites, 1,919 tests passed
- Focused controller: 57 tests passed
- System tests: 13 runs, 86 assertions, 0 failures
- `npm run build` passed
- `bin/complexity_check` passed
- `git diff --check` passed

@codex review

sh1nj1:
## Review follow-up 5: superseding reload and stale hover response

Both P2 findings were confirmed and fixed in `a531d3aab`.

1. **Preserved state across superseding reloads**: a preserved reload now reuses the pending state captured from the rendered rows instead of recapturing expansion and focus from the loading placeholder. A non-preserving load still clears the state.
2. **Invalidated in-flight hover loads before drop**: hover expansion carries a generation check. Drop, dragleave, target changes, and dragend invalidate the generation, so a response created before the move cannot replace the post-drop DOM.

Regression coverage includes overlapping preserved loads, drop during a pending lazy load, dragleave during a pending lazy load, and the inactive cleanup path. Added lines and conditional branches are covered.

Verification:

- Jest: 143 suites, 1,923 tests passed
- System tests: 13 runs, 86 assertions, 0 failures
- `npm run build` passed
- `bin/complexity_check` passed
- `./bin/rubocop -a` passed (1,345 files)
- `git diff --check` passed

@codex review

sh1nj1:
## Review follow-up 6: workspace hover invalidation and async view restoration

Both P2 findings were confirmed and fixed in fb2dddf54.

1. **Workspace hover invalidation**: preview cleanup now cancels and generation-invalidates the workspace branch request on target changes, dragleave, drop, and dragend. A response that survives abort cannot splice stale children into the post-drop tree. Successful child drops explicitly retain the target expansion for the authoritative reload; rejected drops do not leave stale expansion state.
2. **View state lifetime**: preserved state now remains pending until asynchronous ancestor restoration settles. Each load owns a restoration generation, so a superseding load reuses the original expansion/focus snapshot while obsolete restoration stops before mutating the replacement DOM or clearing that snapshot.

Verification:

- Jest: 143 suites, 1,927 tests passed
- Focused regression suites: 121 tests passed
- Added executable lines covered
- System tests: 13 runs, 86 assertions, 0 failures
- npm run build passed
- bin/complexity_check passed
- ./bin/rubocop -a passed (1,345 files)
- Diff whitespace check passed with the repository's JavaScript-compatible whitespace rules

@codex review

sh1nj1:
## Codex review clean on `fb2dddf543` — remaining blocker is the rebase

No code changes this round; the review came back with no findings. What is left is the `CONFLICTING` state, so I ran the rebase to completion as a simulation and verified it end to end. Reporting the result here so the merge is a replay, not an exploration.

### Why it conflicts

The branch carries its own pre-merge copies of the two upstream PRs:

| carried commit | upstream | status |
|---|---|---|
| `c918ebe36` refactor: extract shared drag-and-drop data layer | #1630 → `faf124a4b` | merged, **content evolved during its review** |
| `1835f6562` feat(dnd): add DOM-independent creative move command | #1629 → `3b3b5d3cb` | still open, **has one extra commit** (`fix(dnd): never report a login redirect as a landed move`) |

Both upstream versions moved after this branch snapshotted them, so the copies no longer replay as no-ops. Rebasing commit-by-commit conflicts on the very first T3 commit and each intermediate resolution has to anticipate the eight that follow. Since the merge is a squash anyway, applying the T3 delta (`c918ebe36..fb2dddf54`) as one unit is both simpler and correct.

### Simulation

Base built as `feature/dnd-expansion` + a squash of #1629, then the T3 delta applied three-way. That collapses to **4 conflicts in 2 files**, all resolvable as unions:

- `event_handlers.js` imports — keep `dragState` (used by #1630's `getLastDragOverPosition` calls) alongside T3's `runMoveWithDomRecovery` / `MOVE_STATUSES` / `reportPartialMove` / `expandBranchWithChildren`. `sendNewOrder` and `sendLinkedCreative` drop out: T3 moved those calls into `move_command.js` and nothing in the merged file references them.
- `handleDragOver` and `handleDragLeave` — `setLastDragOverRow(null)` (base) and `clearHoverExpand()` (T3) are independent; `clearHoverExpand` does not touch the last-row state, so both are needed.
- `move_command.js` imports — union of `isAuthenticationRedirect` (used at the unauthenticated check, from #1629) and `serverErrorMessage` (used by the localized partial-failure fallback, from T3).

Three test adjustments follow mechanically:

- `creative_tree_drop.test.js` and `topic_move_released_agent.test.js` mock `lib/api/drag_drop` and must now export `isAuthenticationRedirect`, otherwise the suites fail to load.
- `event_handlers_data.test.js` asserted `effectAllowed === 'move'`. That is the assertion the round-2 fix deliberately changed; it becomes `'copyMove'`.

### One real semantic change, worth flagging

`workspace_tree_adapter.test.js` had a case asserting a drop **completes** when the envelope carries no originating tree. That premise no longer holds. #1630's final `readCreative` rejects a creative envelope without `treeId`:

```js
if (!isPayload(parsed) || !parsed.creativeId || !parsed.treeId) return null;
```

The unified `DND_MIME_TYPE` envelope that used to accept a payload without `treeId` was removed in #1630's own review cycle, and the surviving legacy reader requires it. Both production writers always set one — `event_handlers.js` from `state.treeId`, `workspace_tree_adapter.js` from `row.id` — so the case is unreachable rather than broken. The test now asserts the new contract: such a drop is ignored, `execute` is never called, and the tree is unchanged.

Also removed in the resolution: `renderCreativeTree` and `dispatchCreativeTreeUpdated` are imported in `event_handlers.js` but unused — leftovers from when the lazy loader moved into `branch_expansion.js`. They are dead on `fb2dddf54` too.

### Verification of the simulated rebase

- Jest: **144 suites, 1943 tests, 0 failures** (up from 1927 — the base suites join)
- `workspace_tree_drag_drop_system_test.rb`: 3 runs, 15 assertions, 0 failures
- `drag_drop_system_test.rb` + `workspace_tree_drawer_test.rb`: 10 runs, 71 assertions, 0 failures
- `npm run build` clean
- `bin/complexity_check`: ratchet OK, none grew

### Merge order

#1629 is `MERGEABLE` / `CLEAN` with all eight checks green. It has to land first — rebasing this branch before then would leave the stale T2 copy behind and re-conflict the moment #1629 merges. Once it lands I will replay the resolution above against the real squash commit and force-push.

Not touching the branch until then.

sh1nj1:
## Verification of `13d52dcc6` — approve after two fixes

Verified the native T3-only rebuild against the real `faf124a4b` (T1) / `46e7d8ed3` (T2) stack. The structure is right and every review fix from the `fb2dddf54` line survived the rewrite. Two defects, both in the Ruby test layer, and both fixed — commit `9d7bde913` on `fix/t3-system-test-async-driver`, ready to cherry-pick.

### 1. P1 — the two rejection tests never ran

`drag_between_workspace_rows` calls `page.execute_async_script`. `Capybara::Session` forwards `execute_script` and `evaluate_script` but has no `execute_async_script`, so both tests errored before dispatching a drag:

```
NoMethodError: undefined method 'execute_async_script' for an instance of Capybara::Session
    workspace_tree_drag_drop_system_test.rb:129:in 'drag_between_workspace_rows'
```

That is exactly the coverage this PR asks to be validated — visible-descendant rejection and 403 placement preservation. On `13d52dcc6` the file reports **5 runs, 24 assertions, 2 errors**.

The fix routes the variant script through the same driver-level runner and script timeout `html5_drag_by_offset` already uses, extracted as `html5_drag_async(source, target, x, y, step_delay:, script:)`. `html5_drag_by_offset` keeps its signature, so `system_drag_helper.rb` is untouched.

After the fix: **5 runs, 35 assertions, 0 failures, 0 errors.** Both rejection tests pass — the implementation was correct, only the driver call was wrong.

### 2. P3 — the failing `lint` check

`Layout/EmptyLinesAroundClassBody` at `workspace_tree_drag_drop_system_test.rb:155`. Single autocorrectable offense; it is the whole `lint` failure. Fixed in the same commit. Targeted RuboCop is clean afterwards.

### Structure

The stack is now `main 5ca92af7c` → `faf124a4b` (T1) → `46e7d8ed3` (T2) → `13d52dcc6` (T3), so `MERGEABLE` is real rather than a resolved snapshot. T4–T7 code is absent. `createDragDropRegistry` is a genuine improvement over the previous adapter pair: cross-registry ownership resolution (innermost element wins), preview lifecycle with cleanup handles, and hit intent as data.

Two structural points I checked because they are the kind of thing this refactor breaks silently, and both hold:

- `creative_tree_row.js` drops its `data-action` drag wiring in favour of document-level listeners. That only works because `CreativeTreeRow#createRenderRoot` returns `this` — light DOM, so `event.target.closest('.creative-tree')` still resolves. Under shadow DOM the retarget would have broken every central-tree drag.
- The workspace registry is rooted at `treeTarget`, not `document`, and its handlers `stopPropagation()`. The central registry's document-level listeners therefore never see workspace drags. Cleanup across the pair depends entirely on `handleDrop`'s `finally` iterating `liveRegistries` and calling `finishDrag` on each — remove that loop and cross-tree drops leak `is-dragging` and stale `draggedState`.

### Review fixes carried through the rewrite

All seven confirmed present in the new code, not just in name:

| fix | where |
|---|---|
| lazy children before expanding | `branch_expansion.js` with `isCurrent` generation guard |
| empty lazy branch collapses | `branch_expansion.js`, and `collapseEmptyBranch` for the workspace path |
| response generation guard | `expandBranchForDrag` checks both `loadRequestId` and `dragExpandGeneration` |
| aborted branch id cleanup | `cancelDragExpansion` drops an id absent from `committedExpandedCreativeIds` |
| rendered row, not the id set | `expandBranchForDrag` reads `hoveredItem.dataset.expanded` |
| view state across superseding reloads | `_pendingViewState` reuse plus `_viewRestoreGeneration` in `tree_controller.js` |
| localized partial-failure notice | `move_feedback.js` only shows `serverMessage`, else the EN/KO fallback |

On the wording of `drag_drop.partial_failure`: it says "links", which reads wrong for a move — but `move_command.js` documents `move` as atomic (`reorder_multiple` authorises everything before mutating), so `partial` is reachable only for `link`. Correct as written.

### Verification

Worktree `plan42-worktree20900v` at `13d52dcc6` + `9d7bde913`.

- Jest: **144 suites, 2002 tests, 0 failures**
- `workspace_tree_drag_drop_system_test.rb`: **5 runs, 35 assertions, 0 failures** (24 assertions / 2 errors before the fix)
- `drag_drop_system_test.rb` + `workspace_tree_drawer_test.rb`: **10 runs, 71 assertions, 0 failures**
- `creatives_controller_test.rb` + `workspace_tree_builder_test.rb`: **58 runs, 318 assertions, 0 failures**
- `npm run build` clean
- `bin/complexity_check`: ratchet OK, 395 entities over budget (396 at base), none grew
- Targeted RuboCop: 5 files, no offenses

Worth noting `html5_drag_by_offset`'s own `execute_script` → `execute_async_script` correction in this PR: before it, Ruby resumed before the drop was dispatched, so the three-direction assertions were racing rather than waiting. The three direction tests are meaningful for the first time here.

### Verdict

Approve, conditional on `9d7bde913` landing on this branch. Without it `lint` stays red and the descendant/403 coverage does not execute.

sh1nj1:
## Follow-up verification: source-window bundle synchronization

Verified the P2 fix on `09849f247`.

- Source synchronization consumes every successful `creativeIds` entry, deduplicates identifiers, and retains the scalar `creativeId` fallback for legacy payloads.
- `down` moves replay in reverse while `up` and `child` replay forward, preserving server selection order.
- Link completions do not mutate source rows.
- Missing context/window identity and unusable identifier payloads are ignored.
- No additional defect found in this path.

Verification:

- Jest: 145 suites, 2,025 tests passed
- Focused source-completion suites: 20 tests passed
- T3 system tests: 5 runs, 35 assertions, 0 failures
- `npm run build` passed
- GitHub lint, complexity, schema, scan, JS, and patch coverage checks passed; Rails `test` is still running

The implementation is approved from this review; final merge remains gated on the pending required check.

@codex review

sh1nj1:
## Follow-up 8: retry an unsuccessful active hover expansion

Confirmed the P2 on `09849f247` and fixed it in `5f39f6691` on `fix/t3-hover-expand-retry`.

When a lazy hover load settles without expanding, cleanup now removes the in-flight marker and schedules a fresh 600 ms attempt only if the same connected row is still the active child target and still has its children container. Drop, dragend, target changes, detached DOM, successful expansion, and authoritative empty branches do not retry.

Regression coverage exercises both reported paths:

- a network failure while the registry preview identity remains unchanged;
- pointer leave and re-entry while the obsolete request is still pending.

Verification:

- Jest: 145 suites, 2,027 tests passed
- Focused hover suites: 41 tests passed
- Added statements and conditional branches covered
- Related system tests: 15 runs, 106 assertions, 0 failures
- `npm run build` passed
- `bin/complexity_check` passed
- `./bin/rubocop -a` passed, 1,345 files
- JavaScript-compatible whitespace check passed

The PR branch was not changed. Cherry-pick `5f39f6691` onto `feat/bidirectional-tree-dnd-t3`.

sh1nj1:
Applied the verified hover retry fix in 5f39f66918f3162c7703d35079e47d3be0985734. The exact preserved T3 fix passed 2,027 Jest tests, 15 related system tests / 106 assertions, build, RuboCop and complexity. The integrated candidate abccc65a0e0cc78a24d76b5f27da245d433ba823 passed 154 Jest suites / 2,172 tests and build.

An independent review checked all 18 prior findings against current code and passed eight focused suites / 205 tests. Each has a code fix and regression coverage, including lazy expansion, stale responses, partial-link feedback, copyMove effects, complete cross-window bundle synchronization and unsuccessful hover retries. Resolving the addressed threads; fresh CI remains required before squash merge.

@codex review

sh1nj1:
## The red `test_js` is not this PR

`5f39f6691` shows `test_js` failing, but the Jest run inside that job is green:

```
Test Suites: 145 passed, 145 total
Tests:       2027 passed, 2027 total
```

The job dies two steps later, in `npm audit --production`:

```
js-yaml  4.0.0 - 4.3.1
Severity: high
js-yaml: maxTotalMergeKeys does not limit CPU use for empty merge sources
https://github.com/advisories/GHSA-2883-xcg3-v3hh
1 high severity vulnerability
##[error]Process completed with exit code 1
```

`main` (`5ca92af7c`) carries the same `js-yaml 4.3.1` in its lockfile, so this fails on every branch in the repo — the advisory was published after `main` last ran green. Nothing in T3 touches it.

Worth noting it is a genuine vulnerability rather than audit noise: `js-yaml` is a production dependency and `creative_row_editor.js` parses user-authored YAML in the browser, so the unbounded merge-key expansion is reachable from user input.

Fix is up as **#1637** against `main` — `4.3.2` is the patched `v4-legacy` release, so the `^4` range holds and only the floor moves.

Verified:

- `npm audit --production`: 0 vulnerabilities, was 1 high
- On `main` base: 128 suites, 1749 tests, 0 failures
- Cherry-picked onto this branch at `5f39f6691`: 145 suites, 2027 tests, 0 failures, audit clean
- `npm run build` passes on both

Land #1637 on `main` first, then rerun here — every open PR including the T4–T7 integration work inherits this same red check, and each one carrying its own copy of the lockfile edit would only create conflicts. This branch was not modified.

Remaining on this PR: the Rails `test` job is still running; every other check is green.

sh1nj1:
The latest Jest run passed all 2,027 tests. The test_js job subsequently failed npm audit on newly reported GHSA-2883-xcg3-v3hh (js-yaml). f2c42257 updates only package-lock.json to the compatible patched js-yaml releases, 4.3.2 and nested development copy 3.15.2. Independent clean install, production audit (zero vulnerabilities), 93 relevant tests and build passed. CI is rerunning; no checks were bypassed.

sh1nj1:
All checks green on `69a593874` — `test_js` passes now that `f2c42257e` moved the lockfile to `js-yaml 4.3.2`, and the Rails `test` job finished clean at 15m15s. PR is `MERGEABLE` / `CLEAN`.

#1637 carries the same lockfile entries byte-for-byte against `main`, so the two do not conflict. It is worth landing separately: `main` is still on `4.3.1`, so the T4–T7 integration PRs keep a red `test_js` until it merges. Its own run is green (attempt 1 failed on `CreativeInlineEditTest#test_does_not_duplicate_attachments_when_re-editing_inline_creative`, an unrelated pre-existing system-test flake — that branch touches only `package.json` and `package-lock.json`; attempt 2 passed).

PR #1633, opened 2026-09-08
feat(dnd): bridge touch gestures to registered drop zones

sh1nj1:
Long-press dragging reaches registered sources and drop zones through the same payload and stored hit intent. Taps and swipes remain native before activation; cancellation clears previews and proxies. Lazy expansion refreshes target discovery, edge scrolling follows the pointer, and blocked storage falls back only to the current document's active gesture.

Also preserves the original creative/tree pair in cross-window bundle completion, including non-first dragged rows and partial success. Adapters retain movement commands and DOM recovery.

Validation: 149 Jest suites / 2,106 tests and explicit build passed. Eight new identity regressions fail against the old implementation. Changed registry/touch branches are covered. Vrex accepted the structure; the latest identity delta and CI are under final verification. Rebased onto merged T3 36181e39.

### DIFF
--- a/engines/collavre/app/javascript/creatives/drag_drop/__tests__/creative_tree_drop.test.js
+++ b/engines/collavre/app/javascript/creatives/drag_drop/__tests__/creative_tree_drop.test.js
@@ -389,6 +389,38 @@ describe('right creative tree drop wiring', () => {
     }))
   })
 
+  test.each([
+    [['8', '9'], '9', 'creative-9'],
+    [['8'], '8', null],
+    [['9'], '9', 'creative-9'],
+  ])('pairs the source tree with its successful dragged row for %j', async (succeededIds, creativeId, treeId) => {
+    const consoleError = jest.spyOn(console, 'error').mockImplementation(() => {})
+    window.localStorage.setItem(DRAG_TOKEN_STORAGE_KEY, 'token-under-test')
+    resetDragSessionCache()
+    const setItem = jest.spyOn(Storage.prototype, 'setItem')
+    const completion = jest.fn()
+    window.addEventListener('collavre:creative-drop-complete', completion, { once: true })
+    runMoveWithDomRecovery.mockResolvedValue({
+      status: succeededIds.length === 2 ? 'success' : 'partial', ok: succeededIds.length === 2,
+      succeededIds, failedIds: ['8', '9'].filter(id => !succeededIds.includes(id)), failures: [],
+    })
+    const dataTransfer = transfer()
+    dataTransfer.setData('application/x-collavre-creative', JSON.stringify({
+      creativeId: '9', treeId: 'creative-9', selectedCreativeIds: ['8', '9'],
+      token: 'token-under-test', sourceWindowId: 'other-window',
+    }))
+
+    await handleDrop(dragEvent(tree('1'), dataTransfer))
+
+    expect(completion).toHaveBeenCalledWith(expect.objectContaining({
+      detail: expect.objectContaining({ creativeId, treeId, creativeIds: succeededIds }),
+    }))
+    const signal = setItem.mock.calls.find(([key]) => key === DROP_SIGNAL_STORAGE_KEY)
+    expect(JSON.parse(signal[1])).toEqual(expect.objectContaining({ creativeId, treeId, creativeIds: succeededIds }))
+    setItem.mockRestore()
+    consoleError.mockRestore()
+  })
+
   // A partial view can hand back an envelope whose tree id no longer resolves,
   // so the row identity — not the tree id — has to catch the self drop.
   test('declines an envelope that resolves back onto its own row', async () => {
--- a/engines/collavre/app/javascript/creatives/drag_drop/__tests__/source_window_bundle.test.js
+++ b/engines/collavre/app/javascript/creatives/drag_drop/__tests__/source_window_bundle.test.js
@@ -72,6 +72,25 @@ test.each([
   expect(load).not.toHaveBeenCalled()
 })
 
+test.each([
+  { creativeId: '8', creativeIds: ['7', '8'], treeId: 'creative-8', expected: ['9', '7', '8'] },
+  { creativeId: '7', creativeIds: ['7'], treeId: null, expected: ['8', '9', '7'] },
+])('completion preserves bundle rows while editing defers refresh: %j', detail => {
+  const draft = document.createElement('textarea')
+  draft.value = 'Unsaved target draft'
+  container.lastElementChild.appendChild(draft)
+  document.dispatchEvent(new Event('creative-editing:start'))
+  signal({ ...detail, direction: 'down', targetTreeId: 'creative-9', mode: 'move' })
+  jest.advanceTimersByTime(400)
+  expect(ids()).toEqual(detail.expected)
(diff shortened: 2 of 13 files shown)


sh1nj1:
## T7 review — touch drag parity (`6d90c1c`, six files vs T3 `13d52dc`)

Reviewed the five requested areas. Verified locally at `6d90c1c`: **7 suites / 117 tests pass**, full JS suite **147 suites / 2055 tests pass**, `npm run build` clean. `registry.js` and `touch_bridge.js` are at 100% statements/branches/functions/lines; `touch_drag.js` is 100% lines with only the legacy private guards (186, 236) uncovered — matching the PR body.

### Area-by-area

- **Normal tap/swipe preserved** — Correct. `preserveNativeGestures` removes the `touchstart`/`touchmove` `preventDefault`, `canStart` bails out on non-sources and on editable controls, and `onTap` stays gated to legacy consumers so there is no synthetic click replay. The committed long press still `preventDefault`s `touchend`, which suppresses the compatibility click. `presence_controller.js` is the only legacy consumer and keeps `preserveNativeGestures: false`, so its behavior is unchanged except that a second finger and `touchcancel` now cancel instead of firing a tap — both improvements.
- **Nested zones** — Correct. `getDropTargets()` sorts deepest-first and `targetAtPoint` rejects any candidate that does not contain the real `elementFromPoint` result, so the innermost zone under the finger wins and the ±12px padding cannot produce a false hit through an overlay.
- **Auto-expand target refresh** — The stale `_cachedDropTargets` is gone and targets resolve per update, so newly expanded rows are picked up. **But one defect made auto-expand unreliable in practice — fixed below.**
- **Edge scrolling** — Correct. Distance is clamped to ±`scrollEdge`, the point is bounds-checked against the pane, and the rAF loop is cancelled in `_endDrag`.
- **Cancellation cleanup** — Correct. Escape, `touchcancel`, second finger, and `destroy()` all converge on `finish()` + `_endDrag()`, and the shared per-document bridge finishes a gesture exactly once even when it crosses controller panes.

### Findings

**1. Synthetic `dragleave` carried no `relatedTarget`, breaking hover auto-expand. (fixed)**

`registry.handleDragLeave` guards with `if (event.relatedTarget && activeDrop.element.contains(event.relatedTarget)) return` — that is how a native drag distinguishes moving between two children of one row from actually leaving it. The bridge's `dispatch()` never set `relatedTarget`, so the guard could never fire.

Consequence: any few-pixel drift across a child boundary inside the same row (label span → badge span) tore down the drop preview and, via `event_handlers.handleDragLeave` → `clearHoverExpand()`, reset the 600 ms expansion timer from zero. With normal finger jitter a folder could keep failing to open under a resting finger. Reproduced: preview cleanup ran once and `preview` rebuilt on a 2px intra-row move.

Fixed in `2d3d9ef` by passing the incoming element as `relatedTarget`, matching native `dragleave`. Added a regression test that fails without the change.

**2. `getDropTargets()` walked ancestor chains inside the sort comparator. (fixed)**

Edge scrolling calls `refreshDropTargets()` every animation frame, so this ran ~60×/s during every touch drag, recomputing `depth()` twice per comparison. Fixed in `c32923c` by measuring each depth once before sorting; ordering is unchanged.

**3. The touch drag proxy renders the literal text `1` for single-item drags. (not fixed — needs your call)**

`proxyContent: () => dragImage` returns `null` unless the source called `setDragImage` with a `.drag-bundle-image`, and `_startDrag` then falls back to TouchDragHandler's legacy `${items.length}`. In `singleElement` mode `items` is always `[documentElement]`, so that string is always `"1"`.

Neither live source sets a drag image for a single-item drag: `startWorkspaceDrag` never calls `setDragImage` at all, and `event_handlers.handleDragStart` calls `attachBundleDragImage` only when `selectedCreativeIds.length > 1`. So dragging one row by touch from either tree shows a floating `1` instead of a preview. Reproduced with a source that mirrors `startWorkspaceDrag`:

```
<div class="touch-drag-proxy" style="pointer-events: none; position: fixed; z-index: 10000; ...">1</div>
```

Desktop is unaffected (the browser draws its own snapshot). Two ways to close it, and both land outside T7's six files or change shared visuals, so I did not push a guess:

- **(a) Source-side, preferred:** have `startWorkspaceDrag` and the single-select path of `handleDragStart` call `attachBundleDragImage(event, 1, title)`. Consistent everywhere, but it also replaces the browser's default drag snapshot on desktop — a product decision.
- **(b) Bridge-side:** fall back to `createBundleDragImage(1, label)` when no image was set. Stays inside T7, but the generic bridge would need creative-specific label selectors to find the title.

Worth noting the proxy is otherwise improved here: there is no `.touch-drag-proxy` CSS rule anywhere in the repo, so the new inline `position: fixed; z-index; pointer-events: none` is what makes it track the finger at all.

### Non-blocking notes

- `touch_drag.js` still uses the module-global `document` for the proxy, `document.body.appendChild`, and the `contextmenu` listener, while `registry.js` and `touch_bridge.js` carefully derive `root.ownerDocument`. For a registry rooted in an iframe the proxy would land in the wrong document and the native context menu would not be suppressed. No such registry exists today.
- `_updateDropTarget` treats a `hitTest` returning `false` as a rejection but then compares `hit !== this._currentHit` against a normalized `null`, so a `false`-returning consumer would re-fire `onTargetChange(null)` every frame. The bridge returns `'into' | null`, so nothing hits this today.
- Even with finding 2 fixed, the rAF loop still re-resolves and re-measures the whole target list every frame while the finger is stationary and nothing is scrolling. That is what makes stationary auto-expand work, so I left it alone, but on a large tree it is a per-frame `getBoundingClientRect` sweep worth watching on a real device.
- `touch-action` is not set anywhere. Because the long-press window no longer calls `preventDefault`, a browser that has already latched a pan will ignore the `preventDefault` in `_handleTouchMove` once the drag commits. Within the 10px tolerance this is unlikely, but it is the one thing here I could not verify without a device.

### Fixes

Branch `agent/dnd-t7-review-6d90c1c` (based on `6d90c1c`), two commits to cherry-pick:

- `2d3d9ef` fix(dnd): keep touch previews when the finger drifts inside one zone
- `c32923c` perf(dnd): measure drop target depth once per resolve

After both: 147 suites / 2055 tests pass, build clean, `registry.js` and `touch_bridge.js` still 100% on all metrics. Final branch merge left to the Main owner as requested.

sh1nj1:
Storage-blocked local drag fallback is fixed in a5cf687d54c6955830db4264430aebf81545b19a. It accepts only an active same-document gesture with empty MIME types, preserves rejected foreign payloads, and clears state on completion/cancellation. Independent review and 152 focused tests passed. Full integration: 156 suites / 2,210 tests and build passed.

sh1nj1:
Fixed the source-row pairing finding in 4c95a92481d6738f1dce65748cc6c4990776d3fc. Completion signals retain the original dragged creative/tree pair when that ID succeeded; otherwise they omit the originating tree ID. Eight regressions cover reversed bundle order, partial success, and deferred source-view reloads. The new cases fail against the original implementation and 195 focused DnD tests pass after the fix. Also included the independently verified inline-editor synchronization assertion. Rebased onto the T3 squash 36181e39; fresh CI is running.

PR #1634, opened 2026-09-08
refactor(dnd): migrate comment drag inputs to the shared registry

sh1nj1:
Topics, contexts, selected comments and avatars register drag sources/drop zones with the shared input layer. Domain commands stay in their controllers; shared previews replace duplicated event wiring and highlight classes. This task retains the existing single-creative context/chat command, with bundle behavior in the following T5 PR.

Depends on T3 #1632 and the T7 touch bridge. Each dependency remains separately reviewable; merge in that order.

Validation: 41 comment/preview suites / 738 tests passed, then eight preview/topic tests passed after the default preview fix. Build passed. Vrex review and final Rails regression are pending.

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -50,6 +50,8 @@ module Authentication
     end
 
     def request_authentication
+      return head :unauthorized if request.xhr?
+
       if request.get? && !request.path.start_with?("/inbox") && request.format.html?
         session[:return_to_after_authenticating] = request.url
       end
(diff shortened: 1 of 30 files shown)


sh1nj1:
Single-item adapters now retain payload.creativeId before ids[0] in 3de2bed7f67e0e30ce209c5427e733ff1e26b763. Regression cases cover selection order differing from the active dragged row. 113 focused tests passed. T5 separately expands these commands to all IDs.

PR #1637, opened 2026-09-09
fix(deps): bump js-yaml to 4.3.2 for GHSA-2883-xcg3-v3hh

sh1nj1:
## Why

`npm audit --production` — the last step of the required `test_js` job — now fails on **every** branch in this repository, including `main`:

```
js-yaml  4.0.0 - 4.3.1
Severity: high
js-yaml: maxTotalMergeKeys does not limit CPU use for empty merge sources
https://github.com/advisories/GHSA-2883-xcg3-v3hh
```

The advisory was published after `main` last ran green, so the failure is unrelated to whatever a given PR changes. It is currently the only reason PR #1632 sits at `UNSTABLE` — its Jest run on that same job is `145 suites / 2027 tests, 0 failures`.

## Is it real?

Yes, not just an audit-noise bump. `js-yaml` is a production dependency, and `engines/collavre/app/javascript/modules/creative_row_editor.js` parses user-authored YAML in the browser. Unbounded merge-key expansion is reachable from user input, so the DoS is live rather than build-time only.

## Fix

`4.3.2` is the patched `v4-legacy` release. The `^4` range is kept and only the floor moves, so this is a lockfile bump plus a matching floor in `package.json` — no API change and no other dependency moves.

```
-      "version": "4.3.1",
+      "version": "4.3.2",
```

## Verification

- `npm audit --production`: **found 0 vulnerabilities** (was 1 high)
- Jest on this branch: **128 suites, 1749 tests, 0 failures**
- Cherry-picked onto `feat/bidirectional-tree-dnd-t3` (`5f39f6691`): **145 suites, 2027 tests, 0 failures**, `npm audit --production` clean
- `npm run build` passes on both

## Landing

This should go into `main` first. Every open PR — #1632 and the T4–T7 integration work — inherits the red check until it does, and each of them rebasing the same lockfile edit independently would just create conflicts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "dompurify": "^3.4.13",
         "firebase": "^12.2.1",
         "highlight.js": "^11.11.1",
-        "js-yaml": "^4.1.0",
+        "js-yaml": "^4.3.2",
         "lexical": "^0.38.2",
         "lit": "^3.1.2",
         "marked": "^16.4.1",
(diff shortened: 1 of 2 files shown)


sh1nj1:
Updated after `f2c42257e` landed the equivalent lockfile edit directly on `feat/bidirectional-tree-dnd-t3` (#1632).

This branch now carries the same two lockfile entries byte-for-byte — `js-yaml 4.3.2` for the production dependency and `3.15.2` for the nested dev-only copy under `@istanbuljs/load-nyc-config` — so the two land without conflicting. The extra change here is the `package.json` floor moving from `^4.1.0` to `^4.3.2`, which keeps a fresh resolve off the vulnerable range.

Still worth landing on `main`: `main` (`5ca92af7c`) is still on `4.3.1`, so every open PR other than #1632 keeps a red `test_js` until this merges.

Re-verified on this branch: `npm audit --production` clean, 128 suites / 1749 tests / 0 failures, `npm run build` passes.

sh1nj1:
The dependency change and production audit are verified. This CI attempt passed all 4,823 Rails tests, then failed CreativeProgressToggleKeyboardTest at the immediate database assertion after observing optimistic progress. The existing wait_for_network_idle helper tracks XHR rather than the fetch request used here, so it can finish before the write commits. The dependency diff does not touch this path. Requested one retry of the failed job to distinguish the timing failure; no check was bypassed.

PR #1656, opened 2026-09-10
Extract waiting notice management out of AgentOrchestrator

sh1nj1:
## What

`AgentOrchestrator` owned the "⏳ waiting on the topic slot" notice alongside routing and scheduling: the reason text, the three locks that decide whether a notice is still warranted, the row each defer door writes, and the cleanup that takes it back down.

Two doors defer a dispatch — `AgentOrchestrator#enqueue_jobs` and `AiAgentJob`'s late slot check — and each carried its own copy of the scope decision and the insert. The invariant that they agree (one notice per topic when coalescing is on, 1:1 when it is off) lived in two places and was held by convention.

This moves the responsibility to `Collavre::Orchestration::WaitingNoticeManager` and routes both doors through one `.post`. `AgentOrchestrator` drops 320 lines.

## Ordering preserved

The relocated methods keep their names, bodies and comments. The lock / cancel / dedup ordering they exist to hold is unchanged:

- The scope is still read off the waiter — recorded under the admission lock in `park_waiter` — whenever there is one. The policy fallback is consulted only for a caller with no waiter, and is now passed as a lambda so it is still not evaluated when a waiter answers.
- A shared notice still goes up under `with_deduped_topic_notice` (check and insert on the same `TopicSlot.lock!` the admission takes), a per-deferral notice under `with_live_waiter`, and a `:delayed` notice under neither.
- Cleanup still runs waiter notice → stranded sweep → drained check inside one transaction on the admission lock.

## Behavior change

One, deliberate: the late-admission door now shares the enqueue door's `rescue ActiveRecord::RecordNotSaved` for an error aborted on `:topic`. A notice lost to a topic relocation mid-flight no longer fails the job. Previously only the enqueue door tolerated it, for the same reason.

## Tests

- Existing race-condition tests injected at `AgentOrchestrator.waiting_reason_text` and `.topic_concurrency_notice_exists?`; they now inject at the new owner. Same single method, same call count.
- New `WaitingNoticeManagerTest` covers the unified door directly, including that the policy fallback is **not** consulted when a waiter carries the scope — the invariant the extraction is there to make structural.

Local: `./bin/rubocop -a` clean, targeted orchestration/notice suites green (118 runs), full `bin/rails test` and `bin/rails test:system` reported in a comment below.

Base is `feature/refactor-reduce-dupl` (integration branch), not `main`.

### DIFF
--- a/engines/collavre/app/jobs/collavre/ai_agent_job.rb
+++ b/engines/collavre/app/jobs/collavre/ai_agent_job.rb
@@ -362,7 +362,7 @@ module Collavre
       if waiter.waiting_notice_scope == Comment::WAITING_NOTICE_TOPIC
         Orchestration::TaskCoalescer.coalesce!(waiter)
       end
-      Orchestration::AgentOrchestrator.post_topic_concurrency_notice(
+      Orchestration::WaitingNoticeManager.post_topic_concurrency_notice(
         creative_id, topic_id, context, agent: agent, waiter: waiter
       )
 
--- a/engines/collavre/app/models/collavre/comment.rb
+++ b/engines/collavre/app/models/collavre/comment.rb
@@ -4,7 +4,7 @@ module Collavre
 
     STREAMING_PLACEHOLDER_CONTENT = "..."
     # Authorless "⏳" waiting-notice system messages posted when an agent is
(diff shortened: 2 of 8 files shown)


sh1nj1:
@codex review

sh1nj1:
@codex review

PR #1677, opened 2026-09-11
fix: restore full-width mobile workspace content

sh1nj1:
Remove the 44px left gutter introduced by #1674 so mobile workspace content uses the full screen width again. Keep the tree toggle floating at the bottom with its existing safe-area offset and chat-sheet stacking behavior.

The floating handle can temporarily cover a row while scrolling. Users can scroll that row above the handle; this preserves the requested full-width layout without adding a dedicated side or bottom bar. Replace the former gutter-specific assertion with coverage that a row near the end of a long list can scroll above the handle and expand normally.

The no-overlap assertion applies only to the two-panel layout, where the gutter remains. Mobile behavior is covered by the full-width and scroll-clearance regressions.

Validation:
- Drawer system suite: 17 tests, 150 assertions, no failures or errors.
- New full-width regressions at 375px, 430px, and 767px reproduce the previous 44px gap and pass after removal.
- Existing chat-sheet, tablet, and desktop regressions pass.
- Changed-file RuboCop and the complexity ratchet pass.
- Mobile preview checked at 375px: content bounds span 0–375px with no horizontal gutter.

Only the relevant test suite was run, per project workflow.

### DIFF
--- a/engines/collavre/app/assets/stylesheets/collavre/workspace.css
+++ b/engines/collavre/app/assets/stylesheets/collavre/workspace.css
@@ -318,13 +318,6 @@ body.creative-workspace .indent3 {
         max-width: var(--max-width);
     }
 
-    /* Reserve the handle's width outside the scrolling content so row controls
-       never pass underneath it, including in long creative lists. */
-    .creative-workspace-shell > main {
-        margin-left: 2.75rem;
-        width: calc(100% - 2.75rem);
-    }
-
     /* Keep the drawer handle in easy thumb reach on one-panel screens while
        clearing the device's bottom safe area. */
     .creative-workspace-tree-toggle {
--- a/engines/collavre/test/system/workspace_tree_drawer_test.rb
+++ b/engines/collavre/test/system/workspace_tree_drawer_test.rb
@@ -184,29 +184,28 @@ class WorkspaceTreeDrawerTest < ApplicationSystemTestCase
     end
   end
 
-  # Check initial placement too, including the unchanged two-panel gutter.
-  [ MOBILE_WIDTH, TWO_PANEL_WIDTH ].each do |width|
-    test "the closed drawer handle covers no control in the content column at #{width}px" do
-      visit_workspace(width)
-      assert_selector ".creative-workspace-tree-toggle"
+  # Only the two-panel layout reserves a gutter; mobile allows temporary overlap
+  # and relies on the full-width and scroll-clearance tests above.
+  test "the closed drawer handle covers no control in the two-panel content column" do
+    visit_workspace(TWO_PANEL_WIDTH)
+    assert_selector ".creative-workspace-tree-toggle"
 
-      covered = page.evaluate_script(<<~JS)
-        (function () {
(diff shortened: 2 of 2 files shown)


sh1nj1 (commented):
Reviewed `ddee31e95` locally. **No blocking issues** — the change does what was asked and does not reintroduce an unreachable control. Two low-severity notes below.

### Verified

- The diff touches only the `@media (max-width: 767px)` gutter rule. The two-panel (768–1279px) and three-panel bands are byte-identical, and no JS reads the removed `2.75rem`.
- Drawer system suite reproduced independently: **18 runs, 156 assertions, 0 failures, 0 errors**.
- **Reachability measured, not assumed.** With a 35-branch list at 375px / 430px / 767px (37 row controls, 800px viewport), I computed for every control whether a scroll offset exists where it is fully in-viewport and clear of the handle band:
  - `unreachableCount: 0` at all three widths.
  - At `scrollY == maxScroll`, **zero** controls geometrically overlap the handle.
  - Bottom-most row control clears the handle top by **72px** (row bottom 667.5 vs handle top 740).

  So the P2 raised on #1674 — a row control trapped under the handle — does not materialize: every control has a usable scroll position, and the end of the list is never parked under the handle.

### Note 1 (low) — the mobile "covers no control" assertion is now vacuous

`workspace_tree_drawer_test.rb:188` still runs `the closed drawer handle covers no control in the content column` at `MOBILE_WIDTH`. That guarantee is exactly what this PR deliberately gives up at mobile, and the test now passes only because the shared fixture is three creatives tall, so nothing reaches the handle's band at y≈740–784. Anyone who grows the fixture gets a failure reading `the drawer handle overlaps content controls: [...]` for behavior that is now intended. Suggest restricting that case to `TWO_PANEL_WIDTH`, where the gutter genuinely still exists, and letting the new full-width + scroll-clearance tests own the mobile contract.

### Note 2 (low, no change required) — the reachability invariant is implicit

Bottom-most-row reachability holds because the document ends further below the last row than the handle's reserved height (`innerHeight - handle.top` = 60px). Today the slack comes from `main`'s `padding-bottom: 64px`; measured headroom is 72px. That is comfortable, but nothing in the CSS states the dependency, and a future bottom-anchored element (sticky footer, FAB) would consume it silently. A one-line comment next to the `.creative-workspace-tree-toggle` bottom offset tying it to the content's trailing space would keep that link visible. The `branches.last` case does catch a regression here, which is why this is a comment rather than a request.

### Not an issue

`env(safe-area-inset-bottom)` currently resolves to 0 — no layout sets `viewport-fit=cover` — so the handle offset cannot be pushed up by a home indicator. Pre-existing from #1674 and out of scope here.

sh1nj1:
Addressed review note 1 in a4e94dc77. Restricted the closed-handle no-overlap test to TWO_PANEL_WIDTH and documented why mobile is excluded. Mobile continues to be covered by the full-width and scroll-clearance tests at 375px, 430px, and 767px.

Validation: drawer system suite passes with 17 runs, 150 assertions, 0 failures, 0 errors, and 0 skips; changed-file RuboCop and the complexity ratchet also pass. The test count decreased by one because the obsolete mobile no-overlap case was removed.

PR #1678, opened 2026-09-12
feat: browse images across the creative list

sh1nj1:
Opening a creative image previously limited the carousel to that creative. The viewer now includes images across the currently visible creative list, including title images, in DOM order and starts at the clicked image.

The gallery is rebuilt on each open so added, removed, reordered, expanded, and collapsed rows are reflected. Images hidden by collapsed subtrees, rows being edited, or the `hidden` attribute are excluded, along with empty sources, inline editors, avatars, and images outside the list. Unloaded creatives are not fetched.

Visibility filtering applies only when opening the gallery. Hidden images still receive keyboard accessibility attributes so revealing an already-loaded subtree does not require another preparation pass. The editor guard uses the actual `.inline-edit-form-shell` class. Selection mode and chat attachment behavior are preserved.

Validation:
- Focused JavaScript suite: 43 tests passed; controller statement, branch, function, and line coverage all 100%.
- Focused browser suite: 6 tests, 72 assertions passed, including collapse/re-expand keyboard access and hiding/restoring a row during editing.
- An initial browser run encountered a focus race in the existing title keyboard test; the subsequent complete focused run passed.
- RuboCop: 1,407 files passed. Complexity ratchet and asset build passed.

### DIFF
--- a/engines/collavre/app/javascript/controllers/__tests__/creative_image_lightbox_controller.test.js
+++ b/engines/collavre/app/javascript/controllers/__tests__/creative_image_lightbox_controller.test.js
@@ -40,15 +40,19 @@ afterEach(async () => {
   jest.restoreAllMocks()
 })
 
-test('opens the clicked image, blocks navigation and scopes the carousel to its creative', () => {
+test('opens the clicked image, blocks navigation and navigates across the creative list in DOM order', () => {
   const onClick = jest.fn()
   document.querySelector('#row').addEventListener('click', onClick)
   expect(click('#second')).toBe(false)
   expect(onClick).not.toHaveBeenCalled()
   expect(dialog().querySelector('img').src).toBe('http://localhost/second.png')
-  expect(dialog().querySelector('.image-lightbox-counter').textContent).toBe('2 / 2')
+  expect(dialog().querySelector('.image-lightbox-counter').textContent).toBe('2 / 3')
+  click('.image-lightbox-next')
+  expect(dialog().querySelector('img').src).toBe('http://localhost/title.png')
   click('.image-lightbox-next')
   expect(dialog().querySelector('img').src).toBe('http://localhost/first.png')
+  click('.image-lightbox-prev')
+  expect(dialog().querySelector('img').src).toBe('http://localhost/title.png')
   click('.image-lightbox-prev')
   expect(dialog().querySelector('img').src).toBe('http://localhost/second.png')
   expect(dialog().querySelector('.image-lightbox-delete').hidden).toBe(true)
@@ -60,8 +64,8 @@ test('opens the clicked image, blocks navigation and scopes the carousel to its 
 
 test('opens title images and preserves zoom, keyboard navigation and close', () => {
   click('#title')
-  expect(dialog().querySelector('.image-lightbox-counter').textContent).toBe('1 / 1')
-  expect(dialog().querySelector('.image-lightbox-next').style.visibility).toBe('hidden')
+  expect(dialog().querySelector('.image-lightbox-counter').textContent).toBe('3 / 3')
+  expect(dialog().querySelector('.image-lightbox-next').style.visibility).toBe('visible')
   click('.image-lightbox-zoom-in')
   expect(dialog().querySelector('img').style.transform).toContain('scale(1.25)')
   click('.image-lightbox-close')
@@ -83,6 +87,8 @@ test('preserves image descriptions on every carousel entry and clears missing or
   expect(dialog().querySelector('img').alt).toBe(description)
   click('.image-lightbox-prev')
   expect(dialog().querySelector('img').alt).toBe('First')
+  click('.image-lightbox-prev')
+  expect(dialog().querySelector('img').alt).toBe('')
   click('.image-lightbox-prev')
   expect(dialog().querySelector('img').alt).toBe('')
   click('.image-lightbox-prev')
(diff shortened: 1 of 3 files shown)


sh1nj1 (commented):
## Review — list-wide image gallery

Verified locally on `0dfa470` (worktree): `jest creative_image_lightbox_controller.test.js` → **37 passed**. The core change (hoisting the image query out of `open()` into `_listImages()` and scoping it to `this.element` instead of the clicked `.creative-content`) is correct — the controller is mounted on the list container in `creatives/index.html.erb:39`, so DOM order across rows is the list order the user sees. Selection mode (`_selectionActive`), link/keyboard entry (`_imageForEvent`, `openFromKeyboard`) and the chat-attachment subclass are untouched. `<template>` content is not traversed by `querySelectorAll`, so the empty-state template is safe, and editing avatars live outside `.creative-content` (`creative_tree_row.js:260`) so they stay out.

One thing needs a decision before this lands.

### 1. The gallery includes content that is in the DOM but not on screen

`_listImages()` has no visibility filter, and there are two places where the list hides content with an inline `display: none` while leaving it in the DOM:

- **Collapsed subtrees** — `expansion_controller.js:118` sets `childrenDiv.style.display = 'none'` on an already-loaded `.creative-children` container.
- **The row currently being edited** — `creative_row_editor.js:691-692` sets `.creative-row { display: none }` while the inline editor is open.

Probe against this branch (added the collapsed container to the existing fixture, then clicked `#first`):

```
COUNTER = 1 / 4        // expected 1 / 3 — #hidden-img inside
                       // <div class="creative-children" data-expanded="false" style="display:none">
                       // is in the gallery
```

The user-visible effect is that the counter and next/prev walk into images that aren't rendered anywhere on the page, and **the gallery size depends on expansion history rather than on what's on screen**: a fresh load with a collapsed branch gives 3 images, expand-then-collapse on the same visible screen gives 4. Same for a row that happens to be open in the editor.

The PR description frames the contract as "the current rendered creative list", which reads as *visible*, but the implementation is *present in the DOM*. Either is defensible — but pick one deliberately and lock it with a test, because right now no test covers collapsed or being-edited rows, so the behavior is incidental.

If you go with visible-only, filter in `open()` and **not** in `_listImages()`:

```js
_visibleImages() {
  return this._listImages().filter((image) => !this._hiddenInList(image))
}

_hiddenInList(node) {
  for (let el = node; el && el !== this.element; el = el.parentElement) {
    if (el.hidden || el.style?.display === "none") return true
  }
  return false
}
```

Both hiding mechanisms above use inline styles, so this works under jsdom too (`offsetParent` / `checkVisibility()` do not).

Keep `_prepareImages()` on the unfiltered `_listImages()`. If the filter is applied there as well, collapsed images never get `tabIndex`/`role`/`aria-label`, and re-expanding a *loaded* subtree only mutates `style` and `data-expanded` on the container — neither is in the MutationObserver's `attributeFilter: ["src", "alt"]` — so the prep pass is not guaranteed to re-run for the newly revealed images. That would be a silent keyboard-accessibility regression.

### 2. Minor — `.inline-edit-form` is a dead selector

`_listImages()` and `open()` both exclude `.inline-edit-form`, and the new test asserts it by injecting `<div class="inline-edit-form">`. No element in the app ever carries that class: the shell is `#inline-edit-form` / `.inline-edit-form-shell` (`_inline_edit_form.html.erb:1`). The exclusion that actually does the work is `[contenteditable="true"]` (Lexical), plus the fact that neither `#lexical-inline-editor` nor `#markdown-preview` sits inside a `.creative-content`.

Pre-existing, not introduced here, and not a live bug — but the new test reads as coverage for a guard that never fires. Worth either fixing the selector to `.inline-edit-form-shell` or dropping it, so the test asserts something real.

### What I checked and found fine

- Duplicate `src` across rows — indexing is by element identity, so positions stay distinct.
- `index < 0` early return — same filter feeds both the gallery and the lookup, so a clicked, non-excluded image always resolves.
- Delete/download-by-index in the parent controller (`image_lightbox_controller.js:341`) — `downloadAllUrlValue` is unset on the creative list, so those buttons stay hidden and no cross-creative index is ever sent.
- Workspace mode — the left-panel tree lives outside the controller element, so it is correctly not merged into the gallery.

sh1nj1 (commented):
## Re-review — `4f10feef8` — approving

Verified locally on the branch head: `jest creative_image_lightbox_controller.test.js` → **43 passed**, controller coverage **100% stmts / 100% branches / 100% funcs / 100% lines**. Both findings from the previous round are resolved the way I asked, and I swept for the failure modes the fix could still miss.

### 1. Hidden images — fixed, and the filter placement is right

`_hiddenInList()` walks image → `this.element` checking `element.hidden || element.style.display === "none"`, and it is applied **only** in `open()` (line 99), not in `_listImages()`. `_prepareImages()` still runs over the unfiltered list (line 23), so collapsed images keep `tabIndex`/`role`/`aria-label` and are immediately openable when their subtree is re-expanded — the MutationObserver gap I flagged (`attributeFilter: ["src", "alt"]` never fires on a `style`/`data-expanded` change) is avoided. The new test `prepares collapsed images and rebuilds the gallery after expanding and collapsing loaded rows` pins exactly that: it asserts the aria prep on the collapsed child, then expand → `4 / 4`, collapse → `3 / 3`, re-expand → keyboard `Enter`/`Space` → `4 / 4`. That is the regression I was worried about, locked down.

I checked that inline style is sufficient coverage for this app — it is. Both live hiding paths use inline styles (`expansion_controller.js:118`, `creative_row_editor.js:691`), and there is no class-based hiding in the list: no `classList.add('hidden'|'d-none'|…)` anywhere in `engines/collavre/app/javascript/`, and no CSS rule that hides a `.creative-content` ancestor. `visibility: hidden` appears only on toggle buttons, action buttons and popups (`creative_tree_row.js:278,334,496`), never on a row or content wrapper, so not caught ≠ not covered.

One pre-existing oddity found while sweeping, **not** something this PR needs to fix: `creatives.css:191` — `creative-tree-row.show-edit .creative-row { display: none !important }` — never matches. `show-edit` is only ever applied to `.creative-row` itself (`creative_row_swipe.js:9,36`), never to the `creative-tree-row` host. Dead rule; flagging it so nobody later reads it as a hiding path this filter misses.

### 2. `.inline-edit-form-shell` — fixed

`_inline_edit_form.html.erb:1` is `<div id="inline-edit-form" class="inline-edit-form-shell">`, and `creative_row_editor.js:624-626` moves that same shell element into the tree, so the selector now names a live element. It is still defense-in-depth rather than an active guard — the shell contains `#lexical-inline-editor` and `#markdown-preview`, neither of which is a `.creative-content`, so `_listImages()`'s scope selector already excludes its contents — but the guard now keys on something real, which is what I asked for.

### Minor, non-blocking

- **`open()` line 99** — the filter callback parameter `image` shadows the outer `const image` on line 86. `images.indexOf(image)` on line 100 is outside the callback so it resolves to the clicked image and the behavior is correct (the tests prove it), but the shadowing makes that line harder to read than it needs to be. Rename the parameter to `candidate`.
- **`creative_image_lightbox_test.rb`** — `assert_selector ".image-lightbox-counter", text: %r{[13] / 4}` accepts either `1 / 4` or `3 / 4`. The `4` is the part that matters (the row's images rejoin the gallery once editing closes) and that is asserted, but a regex that tolerates two different positions is papering over nondeterministic list ordering after the editor closes. Worth pinning to the expected index so the test would actually fail if ordering regressed.

Neither blocks the merge.

PR #1681, opened 2026-09-12
feat: show agent execution time in activity logs

sh1nj1:
Activity Logs show the latest successful AI task duration in parentheses beside the newest activity entry's relative timestamp, for example `2 minutes ago (1m 23s)`. Chat message timestamps retain their original appearance, and there is no separate execution-time row. The duration is shown once per expanded panel; missing measurements are omitted, and comments without interaction logs do not gain an empty disclosure.

Timing uses the latest attempt's `start` and `completion` events, excluding queue time and previous retries. Inline agents finish timing at model response completion; delegated agents record completion inside the reply-finalization transaction. In-place reviews link the surviving answer to the latest revision's task. The value describes the task attempt, not an individual tool or LLM call. Historical timestamps are not estimated or backfilled. English and Korean support days, hours, minutes, and seconds.

Timing is calculated only when Activity Logs are opened, using a filtered query for event types and timestamps. Comment lists do not preload or query task actions for timing, so prompt/response payloads are not loaded for this feature. The timing note inherits the activity timestamp's metadata styling and includes an accessible label and measurement tooltip.

Validation:
- 27 targeted unit/integration tests (208 assertions) and 1 browser test (14 assertions) passed.
- Query regression tests verify zero task-action SELECTs for initial loads, topic filtering, both pagination directions, and around-comment navigation.
- Tests cover English/Korean placement, multiple logs, zero/missing durations, unchanged message timestamps, absent logs, review finalization, permissions, and escaped log contents.
- Duration service/helper retain 100% line and branch coverage.
- Full RuboCop and complexity ratchet passed; subsequent system-test updates also passed RuboCop.
- Earlier lifecycle validation: 135 unit/integration tests (584 assertions) passed, covering delegated completion, duplicate/invalid replies, rollback, and in-place review responses.

Full suites run in CI per the project's targeted-test push workflow.

### DIFF
--- a/engines/collavre/app/assets/stylesheets/collavre/activity_logs.css
+++ b/engines/collavre/app/assets/stylesheets/collavre/activity_logs.css
@@ -12,12 +12,6 @@
     color: var(--text-muted);
 }
 
-.activity-log-duration {
-    margin: var(--space-2) 0;
-    color: var(--text-muted);
-    font-size: 0.9em;
-}
-
 /* List container */
 .activity-log-list {
     display: flex;
--- a/engines/collavre/app/assets/stylesheets/collavre/activity_logs.css
+++ b/engines/collavre/app/assets/stylesheets/collavre/activity_logs.css
@@ -12,6 +12,12 @@
     color: var(--text-muted);
 }
 
+.activity-log-duration {
+    margin: var(--space-2) 0;
+    color: var(--text-muted);
+    font-size: 0.9em;
+}
+
 /* List container */
 .activity-log-list {
     display: flex;
--- a/engines/collavre/app/assets/stylesheets/collavre/comments_popup.css
+++ b/engines/collavre/app/assets/stylesheets/collavre/comments_popup.css
@@ -2397,3 +2397,8 @@ li:hover .topic-create-option {
     cursor: pointer;
     font-size: 1.2rem;
 }
+
+.comment-execution-time {
+  color: var(--text-muted);
+  font-size: 0.9em;
+}
--- a/engines/collavre/app/assets/stylesheets/collavre/comments_popup.css
+++ b/engines/collavre/app/assets/stylesheets/collavre/comments_popup.css
@@ -2397,8 +2397,3 @@ li:hover .topic-create-option {
     cursor: pointer;
     font-size: 1.2rem;
 }
-
-.comment-execution-time {
-  color: var(--text-muted);
-  font-size: 0.9em;
-}
--- a/engines/collavre/app/controllers/collavre/comments/activity_logs_controller.rb
+++ b/engines/collavre/app/controllers/collavre/comments/activity_logs_controller.rb
@@ -1,6 +1,8 @@
 module Collavre
   module Comments
     class ActivityLogsController < ApplicationController
+      helper Collavre::ActivityLogsHelper
+
       before_action :set_comment
       before_action :ensure_permission
 
--- a/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
+++ b/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
@@ -11,6 +11,11 @@ module Collavre
       format_execution_time(seconds)
     end
 
+    def inline_task_execution_time(task)
+      seconds = TaskExecutionTime.seconds(task) if task
+      format_execution_time(seconds) if seconds
+    end
+
     def format_execution_time(seconds)
       days, remainder = seconds.round.divmod(86400)
       hours, remainder = remainder.divmod(3600)
--- a/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
+++ b/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
@@ -2,15 +2,6 @@
 
 module Collavre
   module ActivityLogsHelper
-    def task_execution_time(task)
-      return t("collavre.comments.activity_logs.in_progress") if task.active?
-
-      seconds = TaskExecutionTime.seconds(task)
-      return t("collavre.comments.activity_logs.unavailable") unless seconds
-
-      format_execution_time(seconds)
-    end
-
     def inline_task_execution_time(task)
       seconds = TaskExecutionTime.seconds(task) if task
       format_execution_time(seconds) if seconds
--- a/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
+++ b/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
@@ -12,9 +12,10 @@ module Collavre
     end
 
     def format_execution_time(seconds)
-      hours, remainder = seconds.round.divmod(3600)
+      days, remainder = seconds.round.divmod(86400)
+      hours, remainder = remainder.divmod(3600)
       minutes, seconds = remainder.divmod(60)
-      units = { hours: hours, minutes: minutes, seconds: seconds }
+      units = { days: days, hours: hours, minutes: minutes, seconds: seconds }
       units.filter_map do |unit, count|
         next if count.zero? && (unit != :seconds || units.values.any?(&:positive?))
 
--- a/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
+++ b/engines/collavre/app/helpers/collavre/activity_logs_helper.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Collavre
+  module ActivityLogsHelper
+    def task_execution_time(task)
+      return t("collavre.comments.activity_logs.in_progress") if task.active?
+
+      seconds = TaskExecutionTime.seconds(task)
+      return t("collavre.comments.activity_logs.unavailable") unless seconds
+
+      format_execution_time(seconds)
+    end
+
+    def format_execution_time(seconds)
+      hours, remainder = seconds.round.divmod(3600)
+      minutes, seconds = remainder.divmod(60)
+      units = { hours: hours, minutes: minutes, seconds: seconds }
+      units.filter_map do |unit, count|
+        next if count.zero? && (unit != :seconds || units.values.any?(&:positive?))
+
+        t("collavre.comments.activity_logs.#{unit}", count: count)
+      end.join(" ")
+    end
+  end
+end
--- a/engines/collavre/app/helpers/collavre/comments_helper.rb
+++ b/engines/collavre/app/helpers/collavre/comments_helper.rb
@@ -1,7 +1,5 @@
 module Collavre
   module CommentsHelper
-    include ActivityLogsHelper
-
     def formatted_comment_action(comment)
       JSON.pretty_generate(JSON.parse(comment.action))
     rescue JSON::ParserError, TypeError
--- a/engines/collavre/app/helpers/collavre/comments_helper.rb
+++ b/engines/collavre/app/helpers/collavre/comments_helper.rb
@@ -1,5 +1,7 @@
 module Collavre
   module CommentsHelper
+    include ActivityLogsHelper
+
     def formatted_comment_action(comment)
       JSON.pretty_generate(JSON.parse(comment.action))
     rescue JSON::ParserError, TypeError
--- a/engines/collavre/app/services/collavre/ai_agent/response_finalizer.rb
+++ b/engines/collavre/app/services/collavre/ai_agent/response_finalizer.rb
@@ -48,6 +48,8 @@ module Collavre
           reassociate_activity_logs(@reply_comment, surviving)
           # Run row survives via ON DELETE SET NULL, so no run-id bookkeeping here.
           @reply_comment.destroy!
+          # The visible reply belongs to the latest revision's task.
+          surviving.update!(task: @task)
           surviving
         else
           # Normal comment workflow
--- a/engines/collavre/app/services/collavre/ai_agent/task_claim_service.rb
+++ b/engines/collavre/app/services/collavre/ai_agent/task_claim_service.rb
@@ -69,6 +69,7 @@ module Collavre
         raise ActiveRecord::RecordNotSaved, "claimed reply task was not running" unless completed == 1
 
         task.reload
+        task.task_actions.create!(action_type: "completion", status: "done", payload: { comment_id: comment.id })
         ActiveRecord.after_all_transactions_commit do
           run_completion_effects(agent, task, comment)
         end
--- a/engines/collavre/app/services/collavre/comments/list_response.rb
+++ b/engines/collavre/app/services/collavre/comments/list_response.rb
@@ -68,7 +68,7 @@ module Collavre
       end
 
       def filtered_scope(visible_scope, topic_id)
-        scope = visible_scope.with_attached_images.includes(:topic, :comment_reactions, :comment_versions, :snapshot_as_result, task: :task_actions)
+        scope = visible_scope.with_attached_images.includes(:topic, :comment_reactions, :comment_versions, :snapshot_as_result, :task)
         scope = search_scope(scope)
         return [ scope.where(topic_id: topic_id).order(id: :desc), false ] if topic_id.present?
 
--- a/engines/collavre/app/services/collavre/comments/list_response.rb
+++ b/engines/collavre/app/services/collavre/comments/list_response.rb
@@ -68,7 +68,7 @@ module Collavre
       end
 
       def filtered_scope(visible_scope, topic_id)
-        scope = visible_scope.with_attached_images.includes(:task, :topic, :comment_reactions, :comment_versions, :snapshot_as_result)
+        scope = visible_scope.with_attached_images.includes(:topic, :comment_reactions, :comment_versions, :snapshot_as_result, task: :task_actions)
         scope = search_scope(scope)
         return [ scope.where(topic_id: topic_id).order(id: :desc), false ] if topic_id.present?
 
--- a/engines/collavre/app/services/collavre/task_execution_time.rb
+++ b/engines/collavre/app/services/collavre/task_execution_time.rb
@@ -2,7 +2,7 @@
 
 module Collavre
   # Measures the latest completed attempt, excluding queue time and earlier retries.
-  # Missing completion events (including external delegation) are not estimated.
+  # Missing completion events are not estimated, including historical delegations.
   class TaskExecutionTime
     def self.seconds(task)
       return unless task.done?
@@ -11,6 +11,7 @@ module Collavre
         .order(created_at: :desc, id: :desc).pluck(:action_type, :created_at)
       return unless events.first&.first == "completion"
 
+      # Each start marks a new attempt, even when the preceding attempt has no terminal event.
       started = events.find { |type, _time| type == "start" }
       return unless started
 
--- a/engines/collavre/app/services/collavre/task_execution_time.rb
+++ b/engines/collavre/app/services/collavre/task_execution_time.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Collavre
+  # Measures the latest completed attempt, excluding queue time and earlier retries.
+  # Missing completion events (including external delegation) are not estimated.
+  class TaskExecutionTime
+    def self.seconds(task)
+      return unless task.done?
+
+      events = task.task_actions.where(action_type: %w[start completion])
+        .order(created_at: :desc, id: :desc).pluck(:action_type, :created_at)
+      return unless events.first&.first == "completion"
+
+      started = events.find { |type, _time| type == "start" }
+      return unless started
+
+      events.first.last - started.last
+    end
+  end
+end
--- a/engines/collavre/app/services/collavre/task_execution_time.rb
+++ b/engines/collavre/app/services/collavre/task_execution_time.rb
@@ -7,8 +7,7 @@ module Collavre
     def self.seconds(task)
       return unless task.done?
 
-      events = task.task_actions.where(action_type: %w[start completion])
-        .order(created_at: :desc, id: :desc).pluck(:action_type, :created_at)
+      events = execution_events(task)
       return unless events.first&.first == "completion"
 
       # Each start marks a new attempt, even when the preceding attempt has no terminal event.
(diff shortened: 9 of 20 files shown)


sh1nj1:
Reviewed the measurement basis and the activity-log rendering by tracing the actual task lifecycle. The core calculation is sound; one issue blocks merge because it regresses the most common reply type in this workspace.

## Blocking

**Delegated (Claude Channel) replies now render an empty Activity Logs panel.**

`_comment.html.erb` widened the disclosure condition to `comment.task_id.present? || comment.activity_logs.exists?`. `TaskClaimService#link_reply` sets `task_id` on every Claude Channel reply comment, so the panel now appears on all of them. But those tasks can never produce a duration:

- `AiAgentService#delegate_to_claude_channel` logs only `start` and `delegated` (ai_agent_service.rb:57,71).
- The `delegated -> done` transition happens in `TaskClaimService#finalize` via `update_all`, which writes no `task_action`.
- `completion` is written in exactly one place repo-wide — ai_agent_service.rb:124, the inline LLM path.

So `TaskExecutionTime.seconds` returns nil, and the whole panel content is `Execution time: Unavailable` + `No activity logs available.` Before this PR those comments had no panel at all. Net: a new always-empty disclosure on the dominant agent path, and the durations the feature was asked for are the ones it cannot show.

This does not require estimating from `updated_at`. `TaskClaimService#finalize` holds the exact completion instant — log a `completion` action there (it already reloads the task), or have `TaskExecutionTime` fall back to `task.reply_comment.created_at`, which is a recorded event, not an inference. Either way the delegated path becomes measurable and the panel earns its place.

## Non-blocking

1. **`completion` is logged before finalization.** ai_agent_service.rb:124 fires ahead of `finalize_response` and `dispatch_a2a`, so the number excludes reply persistence and A2A dispatch. Acceptable as "model execution time", but `duration_hint` says "from start to completion of the latest attempt", which reads as end-to-end. Either measure to `reply_created` or sharpen the hint.

2. **`failed` / `cancelled` / `escalated` show Unavailable.** These are the runs whose elapsed time matters most — a `TurnDeadlineError` writes a `failed` action with a real timestamp, so `start -> failed` is measurable. Consider keying on a terminal-event set instead of gating on `done?`.

3. **Consecutive `start` records under-report.** With `[start, start, completion]` (a retry that died before logging anything), `events.find { |type, _| type == "start" }` takes the newest start. Unlikely in practice; worth a comment at minimum.

4. **`.activity-log-duration` has no CSS.** Every sibling class is styled in `activity_logs.css`; this one renders with default `<p>` margins.

5. **No day unit.** A long delegated task reads `25h`. Minor, but relevant once (1) lands and long-running delegations become measurable.

## Verified correct

- Ordering by `created_at desc, id desc` plus the "newest event must be `completion`" guard makes negative durations structurally impossible and correctly refuses to pair an earlier completion with a later attempt. The equal-timestamp and out-of-order-insert tests cover this well.
- `active?` includes `pending_approval` and `queued`, so approval waits read `In progress` rather than a misleading number.
- The partial is rendered only by `ActivityLogsController`, so the `helper` declaration is sufficient and the permission check is untouched.
- i18n nesting matches the existing `collavre.comments.*` scope; en/ko parity is complete, and folding the previously hardcoded `Loading...` / `No activity logs available.` / `ago` into i18n is a good cleanup.

sh1nj1:
Addressed the delegated-reply blocker and the automated review finding in `84218711f`.

- `TaskClaimService#finalize` now records `completion` inside the existing reply transaction. Duplicate replies cannot record a second completion, and a completion-write failure rolls back the claim and reply.
- In-place review finalization links the surviving answer to the latest revision's task. Both the disclosure and its timing remain available after placeholder deletion, with or without activity logs. Earlier task events are retained.
- Clarified the English/Korean measurement hint, added timing-row CSS and day units, and documented/tested that consecutive `start` events mark separate attempts.
- Failed/cancelled/escalated timing remains outside this successful-execution feature; these states still show `Unavailable`. Historical delegations without completion evidence also remain unavailable rather than being estimated.

Validation: 135 unit/integration tests (584 assertions), 1 browser test (6 assertions), RuboCop, and the complexity ratchet passed. The new duration service/helper retain 100% line and branch coverage, and both new lifecycle lines are covered.

sh1nj1:
Re-reviewed `84218711f`. The blocker is resolved and the Codex P2 is addressed. I traced both paths and ran the touched tests locally.

## Blocker: resolved

`TaskClaimService#finalize` now writes `completion` after the `running -> done` `update_all`, inside `TaskReplyService`'s transaction. Verified:

- Exactly two places in the repo transition a task to `done` — `AiAgentJob#transition_running_task!` (inline, `completion` at ai_agent_service.rb:124) and `TaskClaimService#finalize` (delegated). Both now emit `completion`, so every successful path is measurable. Claude Channel dispatch never reaches `transition_running_task!` (the `is_claude_channel_agent` branch holds capacity instead), so there is no double-write.
- `completed == 1` on the `status: "running"` predicate makes a second completion impossible; the added conflict/rollback/non-running tests pin all three.
- `start` is logged by `delegate_to_claude_channel` before `ClaudeChannelAdapter#deliver`, so `start -> completion` spans the whole delegation.

Putting the write inside the reply transaction does mean an analytics insert can now fail a Claude reply. `TaskAction` only validates `action_type`/`status` presence and the payload is a two-key hash, so the realistic failure set is DB-level — and in that case rolling the claim back to `delegated` so the reply can be retried is the right outcome.

## Codex P2 (in-place review): resolved

`surviving.update!(task: @task)` runs after `@reply_comment.destroy!`, so the task never has two comments on `task_id` — the ambiguity `Comment#topic_blocking_task` warns about (comment.rb:104) is avoided. `ReviewHandler#eligible?` requires `quoted_comment.user_id == @agent.id`, so only agent-authored comments ever acquire a `task_id` this way; no human comment grows an activity panel. `surviving` is the same memoized instance `ReviewHandler#handle` already wrote, so the partial update touches `task_id` only, and `link_preview_enqueue_required?` gates on `saved_change_to_content?` so no spurious preview job. Side benefit: `broadcast_stop_button_removal` now has a comment to replace in the review flow, where it previously returned early.

## Verified locally

`task_claim_service_test`, `task_execution_time_test`, `activity_logs_helper_test` — 16 runs, 58 assertions, 0 failures. `activity_logs_controller_test`, `response_finalizer_test` — 9 runs, 46 assertions, 0 failures. CI: everything green except `test`, still running.

## Two non-blocking follow-ups

1. **`finalized_response?` doesn't recognise the new `completion` action.** `replay_loop_completion.rb:52` checks `%w[reply_created review_updated]` or `reply_comment&.content.present?`. The delegated path writes neither `reply_created` nor `review_updated`, so a delegated task's only evidence is its `reply_comment` — which an in-place review now moves to the newer task. A comment originally produced by delegation, then revised in place after the agent is pointed at an inline model, leaves the old task reading as `unsuccessful_loop_response?`. Narrow, and it needs an agent reconfiguration to reach, but adding `completion` to that action list closes it in one token and makes the predicate independent of who currently owns `task_id`.

2. **Duration and log list now describe different spans on a reviewed comment.** `reassociate_activity_logs` merges the revision's logs into the comment that already holds the original run's logs, while the duration covers only the latest attempt. A comment from a 2-minute first run plus a 20-second revision reads `Execution time: 20s` above logs from both. The hint says "latest attempt", so it is defensible; scoping the list to the latest attempt, or labelling the revision logs, would make the panel self-consistent.

Neither blocks. Good to land once `test` goes green.

sh1nj1:
Reviewed `c6d77f43a` (inline duration beside the comment timestamp). I ran the touched tests, probed the query count, and checked the rendering on a local preview with a real completed task. It works and reads well — `GitHub PR Analyzer · 19분 전 (2초)`. Two things I'd fix before merge.

## Verified working

- Renders correctly in both locales; `inline_duration` is present in en and ko.
- No extra query for the task itself. `_comment.html.erb:5` already touches `comment.task` for `engine_login`, and `ListResponse#filtered_scope` preloads `:task`.
- The streaming path is clean. `ResponseStreamer` re-renders the partial every 100 ms, but the task is `running` there, so `task_execution_time` returns `In progress` from `active?` without ever reaching `TaskExecutionTime`, and the task instance is memoized on the comment.
- Live transitions work. `became_terminal?` fires `broadcast_stop_button_removal`, which re-renders the comment partial, so `(진행 중)` flips to a real number without a reload.
- 17 tests pass locally (`execution_time_test`, `task_execution_time_test`, `activity_logs_helper_test`), 64 assertions, 0 failures.

## Blocking

**1. N+1: one `task_actions` query per comment on every comments-list load.**

Measured, not inferred. With 8 comments carrying tasks, `GET /creatives/:id/comments` issues exactly 8 of these:

```sql
SELECT "task_actions"."action_type", "task_actions"."created_at" FROM "task_actions"
WHERE "task_actions"."task_id" = ? AND "task_actions"."action_type" IN (?, ?)
ORDER BY "task_actions"."created_at" DESC, "task_actions"."id" DESC
```

`filtered_scope` preloads `:task`, but `TaskExecutionTime.seconds` calls `task.task_actions.where(...)`, so the preload cannot be used — the `.where` forces a fresh query regardless. `LIMIT` is 20, and the list refetches on every topic switch and every infinite-scroll page, so this is up to 20 extra round trips per fetch. Before this commit `TaskExecutionTime` ran once, on demand, for the single panel a user chose to expand.

Fix: preload `task: :task_actions` in `filtered_scope` and have `TaskExecutionTime` filter in Ruby when `task.task_actions.loaded?`, or batch one grouped query over the page's task ids and pass it down as a local.

**2. `(Unavailable)` is now permanently visible on comments it does not help.**

The commit's own test pins this: `assert_select ".comment-execution-time", text: "(Unavailable)"`. That state fires for every `failed` / `cancelled` / `escalated` task, and for every delegated task that finished before this ships — `TaskClaimService#finalize`'s `completion` write is new in `84218711f` and the diff carries no backfill. Delegation is the dominant reply path here, so a large share of existing agent replies get `(측정 불가)` welded to the timestamp.

In the disclosure panel `Unavailable` was fine — the user opted in by expanding, and the surrounding text explains it. Inline it is an answer to a question nobody asked, on a line the eye always crosses. Suggest rendering the span only when there is a real number (keeping `In progress` is defensible), and leaving `Unavailable` to the panel.

## Non-blocking

**3. `.comment-execution-time` has no CSS.** Same gap as `.activity-log-duration` last round, which this PR fixed with `color: var(--text-muted); font-size: 0.9em`. The inline one did not get the same treatment. Computed style on the preview is `16px / rgb(32,33,35) / 400` — full body weight, identical to the timestamp, while every other metadata sibling (`.comment-topic-link`, `.comment-status-label`) is `0.9em`.

**4. `aria-label` on a bare `<span>` is dropped.** Role `generic` prohibits accessible naming in ARIA 1.2, so conforming AT announces only `(2초)` and the `Execution time:` prefix never lands. The test asserts the attribute is present, which looks like coverage but says nothing about announcement. `role="note"` on the span, or a visually-hidden prefix, makes it real.

**5. `duration_hint` is heavy for a header tooltip.** It is a 25-word sentence, now attached to every agent comment header, sitting next to a `<time>` whose tooltip is a 19-character date. A short inline hint with the full sentence kept in the panel would match the weight of the surroundings.

**6. `escalated` is not in `Task#terminal_status?`.** So `broadcast_stop_button_removal` never fires for it and the comment keeps showing `(진행 중)` until a reload, then reads `(측정 불가)`. Pre-existing (the stop button has the same problem), but this commit gives it a second visible symptom.

## Unrelated, found while testing

`GET /creatives/:id/comments/fullscreen` 500s with `undefined method 'any_filter_active?'`. `CommentsController#fullscreen` renders `collavre/creatives/index`, but `any_filter_active?` is declared `helper_method` on `CreativesController` only (`creatives_controller.rb:638`). Not touched by this PR — worth its own issue.

sh1nj1:
Addressed the inline-timing review findings in `776592deb`:

- Preload task actions with the comment page and calculate from the loaded collection. Eight task-backed replies now issue one task-action query; initial load, topic filtering, and both pagination directions have query-count regression coverage. Both calculation paths pass the same retry, missing-event, equal-timestamp, and out-of-order cases.
- Render inline parentheses only for measured durations (including zero). Active, unsuccessful, and historical unmeasurable tasks keep the status inside the activity panel.
- Added muted metadata styling, `role="note"` for the accessible label, and a short localized inline tooltip.

Validation: 26 unit/integration tests (194 assertions), 1 browser test (11 assertions), RuboCop, and complexity ratchet passed. Duration service/helper line and branch coverage remain 100%. The existing preview has been updated.

The pre-existing escalated stop-button behavior and fullscreen error are unchanged; numeric-only inline rendering removes the stale in-progress timing symptom.

sh1nj1:
Re-reviewed `776592deb` (batched timing queries + numeric-only inline rendering). Both items I raised last round are addressed in behaviour, and the Codex P2 on the same subject is addressed too. One thing I'd fix before merge: the batching is correct on query *count* but the preload it uses pulls the entire task-action audit trail, payloads included.

## Resolved

**Inline `(Unavailable)` is gone.** `inline_task_execution_time` returns `nil` unless `TaskExecutionTime.seconds` produced a number, and `format_execution_time(seconds) if seconds` keeps `0` renderable (`0` is truthy in Ruby), so an instant task still shows `(0s)`. The new test walks `running / pending / queued / delegated / pending_approval / failed / cancelled / escalated` plus the missing-`completion` case and asserts `count: 0` for each — that pins exactly the set I was worried about, including historical delegations with no backfill. `Unavailable` stays inside the disclosure where the surrounding text explains it.

**Loaded/unloaded parity holds.** `sort_by { [created_at, id] }.reverse` is a total order (ids are unique) and matches `order(created_at: :desc, id: :desc)` exactly, so the retry, equal-timestamp, and out-of-order-insert cases behave identically on both paths. The test file exercises both.

**Streaming is still query-free.** `TaskExecutionTime.seconds` returns early on `unless task.done?`, so the 100 ms re-render never reaches the events lookup.

**Non-blocking 3/4/5 from last round are done.** `.comment-execution-time` landed in `comments_popup.css` next to `.comment-status-label` / `.comment-topic-link` with the same `0.9em` + `--text-muted` treatment; `role="note"` permits accessible naming under ARIA 1.2, so the `aria-label` is now announced; `inline_duration_hint` is a 4-word tooltip instead of the 25-word sentence.

**Verified locally:** `execution_time_test`, `task_execution_time_test`, `activity_logs_helper_test`, `activity_logs_controller_test` — 26 runs, 194 assertions, 0 failures.

## Blocking: the preload hydrates every task-action payload

`includes(..., task: :task_actions)` issues one query, but it is unscoped and unselected. Measured, not inferred — instrumented `sql.active_record` on a real `GET /creatives/:id/comments` with 8 task-backed comments:

```
SELECT "task_actions".* FROM "task_actions" WHERE "task_actions"."task_id" IN (?,?,?,?,?,?,?,?)
```

`SELECT *`, no `action_type` filter. So every action row for every task on the page is transferred and instantiated, and `task_actions` carries two `json` columns that Rails deserialises per row:

- `prompt_generated` → `{ messages: messages_data[:messages] }` (ai_agent_service.rb:79) — the full prompt, i.e. `creative_context` (the whole creative subtree) plus the entire `chat_history`.
- `completion` → `{ response: @streamer.content }` (ai_agent_service.rb:124) — the full model response.
- `reply_created` → `{ content: @response_content }` (response_finalizer.rb:71,83) — the response text again.
- `review_updated` → includes `content` (review_handler.rb:56).

In my 8-comment probe that is **167,608 bytes** of JSON read and parsed to recover 16 timestamps. And it is not a fixed cost: `prompt_generated` grows with the creative tree and the thread, so it is ~4 KB for a two-message test creative and unbounded for a real topic. This fires on initial load, on every topic switch, and on every infinite-scroll page.

Net, the current fix trades 20 sub-millisecond two-column index reads for one query that can move megabytes plus the JSON parsing. That is a worse shape than the N+1 it replaced, and it is the kind of regression that only shows up on the biggest topics.

A narrow purpose-built association keeps the single query and drops the payload. I prototyped it, measured it, and ran the suite:

```ruby
# task.rb
has_many :execution_actions, -> { where(action_type: %w[start completion]).select(:id, :task_id, :action_type, :created_at) },
         class_name: "Collavre::TaskAction", inverse_of: :task, dependent: nil
```

```ruby
# list_response.rb
includes(:topic, :comment_reactions, :comment_versions, :snapshot_as_result, task: :execution_actions)
```

```ruby
# task_execution_time.rb
def self.execution_events(task)
  actions = task.execution_actions
  return actions.order(created_at: :desc, id: :desc).pluck(:action_type, :created_at) unless actions.loaded?

  actions.sort_by { |action| [ action.created_at, action.id ] }.reverse
    .map { |action| [ action.action_type, action.created_at ] }
end
```

Result on the same probe:

```
SELECT "task_actions"."id", "task_actions"."task_id", "task_actions"."action_type", "task_actions"."created_at"
FROM "task_actions" WHERE "task_actions"."action_type" IN (?,?) AND "task_actions"."task_id" IN (?,?,?,?,?,?,?,?)
```

1 query, **1,363 bytes** instead of 167,608 — a 123x reduction, still 8 inline durations rendered. It also collapses `execution_events` because the scope does the filtering, so the loaded and unloaded branches share one definition instead of duplicating the filter in Ruby. 30 runs / 215 assertions pass (the four files above plus `task_claim_service_test`), RuboCop clean on all three files.

## Non-blocking

1. **`around_comment_id` issues two timing queries.** `comments_around` builds two scopes, so each preloads separately. Structurally fine — two comment queries, two timing queries, no N+1 — but the query-count test covers `{}`, `topic_id`, `after_id`, and `before_id` and not this branch, so "one query per page" is slightly stronger than what is pinned. Worth a case or a note.

2. **`finalized_response?` still does not recognise `completion`.** `replay_loop_completion.rb:51` checks `%w[reply_created review_updated]` or `reply_comment&.content.present?`. The delegated path writes neither of those actions, so a delegated task's only evidence remains its `reply_comment` — which in-place review now moves to the newer task. Carried over from last round, still a one-token fix, still narrow enough that I would not hold the PR for it.

3. **Duration and log list describe different spans on a reviewed comment.** `reassociate_activity_logs` merges the revision's logs into the comment already holding the original run's logs, while the duration covers only the latest attempt, so a 2-minute first run plus a 20-second revision reads `Execution time: 20s` above logs from both. The hint says "latest attempt", so it is defensible; scoping the list or labelling the revision logs would make the panel self-consistent. Also carried over.

Fix the preload width and this is good to land.

sh1nj1:
Re-reviewed `faf23d0ff` (duration moved into the activity-log timestamp, list timing lookup removed). I traced the ActivityLog write paths, diffed the list query shape against the pre-PR base, and ran the touched tests. No blockers. One consequence worth an explicit decision rather than a silent one.

## Resolved

**The list query shape is back to base, not merely narrowed.** `filtered_scope` preloads `:task` — byte-identical to `903a7177a` — and the panel gate is `comment.activity_logs.exists?`, also identical to base (`_comment.html.erb:229` before the PR). So the per-comment `task_actions` query I raised and the Codex P2 on preload width are both gone at the root: nothing in the comment-list render touches `task_actions` at all. The narrow-association fix I prototyped is no longer needed. The query-count test now also covers `around_comment_id`, which closes my open non-blocking #1 from last round.

**Placement matches the instruction.** `.comment-execution-time` is removed from the header, `l(timestamp, format: :chat_timestamp)` is untouched, and the standalone `Execution time: …` row is gone. The number now renders inside `.activity-time` of the first log row. `ActivityLogsController` orders `created_at: :desc`, so `index.zero?` is the newest log — the right anchor for a "latest attempt" figure, and the test pins it via `.activity-log-item:first-child` / "Latest LLM".

**Styling is fine without a new rule.** `.activity-execution-time` has no declaration of its own, but it is nested inside `.activity-time` (`0.85em`, `--text-muted`), so it inherits the muted metadata treatment. Unlike `.comment-execution-time` last round, there is no gap to fill here. `role="note"` + `aria-label` are retained, and the full `duration_hint` sentence is back in the panel, where its weight fits.

**My earlier non-blocking #3 is mitigated.** On a reviewed comment, `reassociate_activity_logs` merges the revision's logs in and the duration covers only the latest attempt — but the number now sits next to the revision's own row instead of floating above a mixed list, so the association reads correctly.

**Verified locally:** `activity_logs_controller_test`, `execution_time_test`, `activity_logs_helper_test`, `task_execution_time_test`, `response_finalizer_test` — 29 runs, 213 assertions, 0 failures.

## Consequence to decide on (not a defect)

**Delegated (Claude Channel) replies now show no duration at all.**

`ActivityLog` is written in exactly two places repo-wide:

- `RubyLlmInteractionLogger.log`, called only from `ai_client.rb:466` — the only writer that sets `comment:`.
- `McpService`'s `after_call`, which passes `comment: nil` and only wraps dynamically registered meta tools (built-in `Mcp::*` tools do not log at all — `ApplicationTool` has no hook).

`delegate_to_claude_channel` never constructs an `AiClient`, so a delegated reply comment has zero `activity_logs`. With the gate back to `comment.activity_logs.exists?`, those comments get no panel, and therefore nowhere to render the duration. The system test pins exactly this: `@historical` asserts `assert_no_selector ".comment-activity-log-block"`.

Net: the `completion` action added in `84218711f` is now unreachable from the UI. It still answers the original question — the timing data is stored and analysable — but the *displayed* feature covers the inline LLM path only, and delegation is the dominant reply path in this workspace.

This follows the requested placement faithfully; there is no "N minutes ago" row on a delegated reply to attach the number to. Flagging it so the trade-off is chosen rather than discovered.

## Non-blocking

1. **`TaskExecutionTime.execution_events`'s `loaded?` branch is now dead.** Grepped for `includes`/`preload`/`eager_load` of `task_actions` across `engines` and `app`: no hits. Only the `pluck` path runs in production. The tests still exercise both, which reads as coverage of a path that no longer exists — collapsing to the pluck would remove the duplicated ordering logic.

2. **Total-attempt duration sits beside a single row's timestamp.** With a tool loop the comment carries several `llm_query` rows, and `(1m 23s)` next to the newest one's "3 minutes ago" invites reading it as that row's own duration. `title`/`aria-label` disambiguate on hover only. A short label, or hanging it off the panel summary, would make it unambiguous at a glance.

3. **`finalized_response?` still does not recognise `completion`** (`replay_loop_completion.rb:51`). Carried over unchanged from the last two rounds; still narrow, still a one-token fix.

CI: 13 checks green, `test` and `Analyze (rust)` still running.

PR #1700, opened 2026-09-18
feat: unify tree drag and drop with design tokens

sh1nj1:
Left and right creative trees now share inset insertion lines, target backgrounds, child arrows, Shift-link badges, and source opacity. Previously the left insertion line also overlapped a host-level border, changing row geometry.

Introduce `--dnd-*` design tokens and shared rules in the existing `collavre/creatives` stylesheet. Color aliases resolve on the themed body so light, dark, and custom themes apply consistently. Existing hosts that link this stylesheet individually keep drag feedback, with fallback values when design tokens are absent. The installation guide now recommends `collavre_stylesheets` for the complete stylesheet set.

Reuse the existing link indicator on the workspace tree and clear source/target feedback on drop, cancellation, and teardown. Production code: +83 / -61 lines (net +22).

Validation:
- D&D Jest regression suites: 334 tests passed on the initial implementation; modified JS files have 100% line coverage. This compatibility follow-up does not change JavaScript.
- Workspace D&D browser suite: 7 tests / 135 assertions passed, including computed-style parity and stable geometry across light, dark, and custom themes, plus an isolated frame loading only `collavre/creatives` without token assets.
- Application helper tests: 7 tests passed on the initial implementation.
- Targeted RuboCop passed after the compatibility fix. Complexity ratchet and asset build passed on the initial implementation.
- Stylelint reports only the six existing hardcoded-color errors in `creatives.css`. The compatibility section has a scoped color-rule exception for fallbacks used by hosts without design tokens.


sh1nj1:
Addressed the P2 finding in `6f751cc6f`.

`collavre_stylesheets` now lives in `Collavre::StylesheetsHelper` (`engines/collavre/lib/collavre/stylesheets_helper.rb`) and an engine initializer mixes it into `ActionView::Base`, so hosts that mount the gem without including `Collavre::ApplicationHelper` get the helper in their own layouts and controller views. `Collavre::ApplicationHelper` still includes it, so existing hosts and engine views are unchanged.

Guard verified: with the initializer removed, the new tests fail (2 failures); with it in place, 12 helper tests and 16 layout-rendering integration tests pass, RuboCop clean.

